PR #16, opened 2016-07-27
FFT now linked to filtered EEG data

sunwangshu:
FFT now happens in dataProcessing.process(), and now it only calculates filtered data instead of raw data.

As a result dataProcessing.process() doesn't take in fftBuff as one of the argument, while dataProcessing_User.process() remains the same (still has reference to fftData).

As a result the function initializeFFTObjects() is no longer in need and was removed.


sunwangshu:
This version works, though now dataprocessing.process() don't take fftBuff but dataProcessing_User.process() still takes fftBuff as one of the argument, and this asymmetricity could cause some confusion. And anyway none of the arrays passed by function are protected. 

If there is an array float[] samplearray composed of all 0 values, and a function void samplefunction( float[] anotherarray) { anotherarray[0] = 1; }. If you call samplefunction(samplearray), samplearray can still be modified. 

So we may plan to remove some of these array passing feature of the functions to simplify the look.

sunwangshu:
ready to go

PR #24, opened 2016-08-14
Playback networking

gabrielibagon:
Networking features added for playback mode.

All Control Panel features for networking needed to be specified as being "live" or "playback" (this is because two different types of boxes/cp5 fields need to exist concurrently, even if only one or neither of them are visible at a given time).

The data to be sent to the network is taken from DataProcessing, where the data is first read from the file. This data is then sent to Networking.pde, where it is sent to the appropriate networking classes and methods.


aj-ptw:
We need to rebase this off master and merge it!

aj-ptw:
Closing this because of #58

PR #58, opened 2016-11-02
Gabe playback networking

aj-ptw:
Add gabe's playback networking!


aj-ptw:
This needs testing

aj-ptw:
@gabrielibagon can this be used to receive data over the network? This is a critical feature needed for wifi and if you already did it... well i want it.

gabrielibagon:
I believe this works for data being played back from the GUI, and sends it through UDP, OSC, LSL. It's old code, though.

PR #122, opened 2017-02-07
Added correct port name for mac users

matthew-groth:
The port name was not correct for me, and I assume other mac users. It was difficult tracking down this problem and fixing it. Hopefully, this change will prevent that issue from happening for some users.

### DIFF
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -157,7 +157,7 @@ OSCSend osc;
 LSLSend lsl;
 
 // Serial output
-String serial_output_portName;  //set in the very beginning of SETUP after testing OS
+String serial_output_portName;  //set later depending on OS
 Serial serial_output;
 int serial_output_baud = 115200; //baud rate from the Arduino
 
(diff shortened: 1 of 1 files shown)


aj-ptw:
What functionality does this add? I've never had mine set correctly lol does it auto find the board? I always though this was for connection to another serial device like Open Bionics. Thanks for the PR!

matthew-groth:
Originally this set the port name to what I thought was probably the correct one for all Mac users but didn't change the way it worked for Linux or Windows users. But after reading your comment, I found a terminal command that does, in fact, find the port name. I only tested this on Mac, so it is set to only run on Mac. After the last commit, it still doesn't change the way it works on Linux or Windows, but I think it would be possible and a good idea to find similar commands for them.

Happy to get involved!

aj-ptw:
Did this fix a problem you were having? I'm still not sure what this patch solves. @biomurph said this was for controlling an auxiliary serial device like an arduino, not an actual OpenBCI board.

matthew-groth:
Before, I was getting the RuntimeException

`    verbosePrint("OpenBCI_GUI.pde: could not open " + serial_output_portName);
`

Now I am not.

I can't say for sure if this is getting me signal though, because the visualizers for Ganglion seem to be disabled.

matthew-groth:
Now this branch is only to stop Mac users from seeing output saying that they are not connecting to the port.

aj-ptw:
Ok i see what is going on here. That line to connect to the serial port should be commented out by default. I'll keep this open but i don't think the proper action is to only do this on Linux/Windows, it's to not do it on **any** operating systems unless some one knows what they are doing and want's to change the source code, the error is misleading and should not be shown to everyday users.

Issue #138, opened 2017-02-25
Modify GUI Headmap Electrode Positions


CelticHarp:
It would be nice the file 'electrode_positions_default.txt' to be editable and readable by the GUI in order to change the default electrode positions that correspond to the GUI headmap.
Maybe at a later improvement we could choose from a picture the positions and the file to be updated automatically with the coordinations.
In any case a documentation of this capability would be essential.

Reported by: compmonks, jussijaatinen, celticharp

Thank you!

compmonks:
yes, it would be nice to have a text file with 10-20 coordinates adjusted for the GUI along with a way to label electrodes with their correct name (Fz, FCz, Oz, O3, ...) instead of their index (1,2,3,...). I guess most users would find it also very useful  and time-saving :)

aj-ptw:
@conorrussomanno wrote some code for this once upon a time

aj-ptw:
closed by #152

emolyzr:
Hi,
thanks for adding the drag and drop option. Unfortunately, this doesn't solve the issue. One can move the electrode positions, but this won't affect the calculation of the scalp distribution in any meaningful way.
![unbenannt](https://user-images.githubusercontent.com/35568338/35920879-6facdc02-0c19-11e8-9022-e1ad00da1f16.png)

In addition, it's not possible to save the configuration, so that upon reloading the GUI one has to move the nods again.

Any help would be highly appreciated.

emolyzr:
As a workaround to solve this issue: 
1.) Just go to ~\OpenBCI_GUI\data 
2.) Open the file electrode_positions_default.txt and change the settings according to your setup (You may wish to make a backup of the original file first). 
3.) When starting the GUI, the head plot widget will always read out this file. 

It would be convenient, to know the coordinates of the other nodes of Ultracortex Headset, however by playing around one should get the right estimates oneself.

yidilozdemir:
I have the latest version of the OpenBCI_GUI, and here I cannot find this location ~\OpenBCI_GUI\data. What is the current fix for the issue described?

PR #152, opened 2017-05-22
Added electrodes drag and drop to the head widget

liqwid:
Also added mouseDrag callback to widget class

If electrodes overlap then electrode with a bigger index is in priority for drag.
On dynamic picture colors update occurs after drop of the electrode

There is a small flaw: when data stream is paused and electrode is dragged the static picture won't update with current data.

#138

### DIFF
--- a/OpenBCI_GUI/Interactivity.pde
+++ b/OpenBCI_GUI/Interactivity.pde
@@ -443,7 +443,16 @@ void parseKeycode(int val) {
   }
 }
 
+void mouseDragged() {
 
+  if (systemMode >= SYSTEMMODE_POSTINIT) {
+
+    //calling mouse dragged inly outside of Control Panel
+    if (controlPanel.isOpen == false) {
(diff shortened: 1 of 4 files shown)


aj-ptw:
Thanks looking forward to trying it out!

aj-ptw:
Not working for me when I try to drag? How does this work?

liqwid:
Plain click and hold. Does the border thicken on hover?

Let me check it out at home.

aj-ptw:
problem is in this chunk of code because `isDraggedElecInsideHead` is returning false when I'm for sure inside the head.

```
  private boolean isDraggedElecInsideHead() {
    int dx = mouseX - circ_x;
    int dy = mouseY - circ_y;
    return dx * dx + dy * dy < (circ_diam - elec_diam) / 4;
  }
```

liqwid:
Made a fix in a pull request

PR #158, opened 2017-06-02
Fixed electrode being inside head condition for head plot widget

liqwid:
Sorry for that mistake, I just forgot that processing IDE can run updates without saving source files.


aj-ptw:
great job.

PR #241, opened 2017-09-05
fps fix

conorrussomanno:


aj-ptw:
ah your fix was different.... i will fix this

PR #284, opened 2017-11-06
Fixing bugs related to makers

gerrievanzyl:
The previous marker code allowed non printable characters to be sent as
the marker values.  This supported markers in range [1:255].  It was
tested using BLE connection.

However, it seems there are inconsistencies in the implementation of
hub.sendCommand(String) in the ElectronHub code.  When using BLE the
string is correctly sent (even when non printable characters were included in the String), however, when using wifi it seems that the  String is UTF-8 encoded.  This introduces additional characters into the stream that at minimum produces unpredictable results - and often restarts the Cyton
board!

For the time being, the best option is to limit markers to the 96
printable ascii characters.  This version of the GUI does that by
mapping 1-96 to ‘ ‘ - ‘~’ .

I will send an associated pull request through for the 32bit firmware


aj-ptw:
> When using BLE the
string is correctly sent (even when non printable characters were included in the String), however, when using wifi it seems that the String is UTF-8 encoded. This introduces additional characters into the stream that at minimum produces unpredictable results - and often restarts the Cyton
board!

Could you send me the exact line? [How to link to a specific line](https://stackoverflow.com/posts/31282559/revisions)

This sounds like a legit issue. 

Thanks!

aj-ptw:
Can you also please provide an example file of using this marker system?

gerrievanzyl:
@aj-pw, I have not looked into the JS code.  In GUI this is the method called by BLE and wifi to send a command to the Electron hub.  I have trapped the string here and they are identical and they both correctly contain the printable and non-printable characters

https://github.com/OpenBCI/OpenBCI_GUI/blob/765111b22fd7e565df59f7208d2d7d0ee1e9a6aa/OpenBCI_GUI/InterfaceHub.pde#L1035-L1053

However, on the receiving side in the 32Bit code: 

https://github.com/OpenBCI/OpenBCI_32bit_Library/blob/a59623e3297f5ec59eb8d350e49e1bd52574c0df/OpenBCI_32bit_Library.cpp#L108-L115

Has the encoded string.  

There is a possibility (I have not explored) that the wifi class in the 32Bit board is doing something strange with the string, but I doubt that.  I think it uses the same character class to read the wifi packet as the bluetooth packet.  

I strongly suspect that the issues is somewhere in ElectronHub.  Hope this helps

aj-ptw:
This has electron hub written all over it

PR #287, opened 2017-11-13
alpha_count and beta_count don't need to be nulled

jooray:
alpha_count and beta_count are local variables, it does not matter if we set them to zero before returning from function, this is unnecessary


aj-ptw:
Thanks @jooray !

Issue #329, opened 2018-05-07
Problem with OpenBCI_GUI 3.3.0


wellton:
## Problem

do not working in Debian GNU/Linux. the program crashes when I click on any menu.

## GUI Version

The version is 3.3.0

See in the picture.

thanx

att
![captura de tela de 2018-05-07 13-00-02](https://user-images.githubusercontent.com/2234645/39711686-c5d58fe8-51f6-11e8-94b4-f6388948c7f2.png)

aj-ptw:
i'm making a linux computer as we speak to work on this.

wellton:
show. Thanks =)

chelobaka:
Same error on Debian. Version 3.2 worked fine.

wjcroft:
AJ, thanks for looking into the 3.3.0 problems.

They are also showing up on Windows as well, in the form of freezing.

http://openbci.com/forum/index.php?p=/discussion/1594/openbci-3-3-0-doesn-t-work-windows-10-64-bit

http://openbci.com/forum/index.php?p=/discussion/comment/8315/#Comment_8315

chrisjz:
I've submitted a fix that stops the freezing in Windows 10, which I think should theoretically also work in Linux too.

Until this change gets verified and accepted, you can grab it from here:
https://github.com/chrisjz/OpenBCI_GUI/archive/development.zip

You will need to run it using Processing though, here are the steps (using the above ZIP file instead of the one mentioned in instructions):
http://docs.openbci.com/OpenBCI%20Software/01-OpenBCI_GUI#the-openbci-gui-running-the-openbci-gui-from-the-processing-ide

aj-ptw:
Just released a new beta with the bug fix from @chrisjz https://github.com/OpenBCI/OpenBCI_GUI/releases/tag/v3.3.1-beta1

cc: @wjcroft

chelobaka:
Original report was about Debian. Version 3.3.1-beta1 doesn't fix it. Now there's an error right after launch:

java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #330, opened 2018-05-11
FIX: Freezing in Windows

chrisjz:
Fixes freezing in Windows from #329. May also work in Linux but I don't have a way to test this.


aj-ptw:
Thanks!!

PR #354, opened 2018-07-06
Development 3.4.0 Alpha 1

retiutut:
Fixing errors that appear in testing while trying to save/load while using Ganglion.

.gitignore all JSON files


retiutut:
When using standalone version compiled from this branch on Mac, it makes a new folder /data/ outside the .app package. Is ok to save UserSettingsFile.json and DefaultSettingsFile.json to /SavedData/ ?

PR #370, opened 2018-08-16
Fix Bug: Playback indexes load properly so playbackSlider can function

retiutut:
Just need to reset indices, hasRepeated, and has_processed. This will work for a future playback widget also!


retiutut:
> so i load a file, exit system, new file, start system....
the playback slider works but it has the old data 

> 
- [x] Fix this issue

retiutut:
This fix has been merged with https://github.com/retiutut/OpenBCI_GUI/tree/playback-widget .

andrewjaykeller:
Thanks @retiutut

Issue #374, opened 2018-08-27
BUG: 3.4.0 not working with WiFi Shield


andrewjaykeller:
## Problem

Running the GUI as standalone i don't get an error and the GUI just freezes. 

Here is the relevant output code

```
Ch 0, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 1
**
Powering up channel 1
Command: !
Hub: sendCommand(char): sending '!'
Ch 1, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 2
**
Powering up channel 2
Command: @
Hub: sendCommand(char): sending '@'
Ch 2, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 3
**
Powering up channel 3
Command: #
Hub: sendCommand(char): sending '#'
Ch 3, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 4
**
Powering up channel 4
Command: $
Hub: sendCommand(char): sending '$'
Ch 4, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 5
**
Powering up channel 5
Command: %
Hub: sendCommand(char): sending '%'
Ch 5, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 6
**
Powering up channel 6
Command: ^
Hub: sendCommand(char): sending '^'
Ch 6, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 7
**
Powering up channel 7
Command: &
Hub: sendCommand(char): sending '&'
Ch 7, Active, x24, Normal, Include, On, Off
OpenBCI_GUI: activating channel 8
**
Powering up channel 8
Command: *
Hub: sendCommand(char): sending '*'
done writing channel.r,set,0,0,24,normal,1,1,0,;

done writing channel.r,set,0,0,24,normal,1,1,0,;

done writing channel.r,set,0,0,24,normal,1,1,0,;
done writing channel.r,set,0,0,24,normal,1,1,0,;

done writing channel.r,set,0,0,24,normal,1,1,0,;

done writing channel.r,set,0,0,24,normal,1,1,0,;
done writing channel.r,set,0,0,24,normal,1,1,0,;

done writing channel.r,set,0,0,24,normal,1,1,0,;

done writing channel.r,set,0,0,24,normal,1,1,0,;

[...]

PR #375, opened 2018-08-27
Fix Bug: Catch error applying channel settings to Cyton #374

retiutut:
Address issue #374


andrewjaykeller:
Thanks @retiutut will test

PR #383, opened 2018-09-23
v4.0.0 Alpha Feature Additions

retiutut:
Address the following issues:

#48 
#55 
and full integration with the solution for #38 

**Add Feature: Playback Widget**
- [x] Time Series update the number of channels when loading successive files using playback widget

- [x] EMG update the number of channels when loading successive files using playback widget 
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/d6bf66dc98b6b914a1ec0dee966da9e44271cb11

- [x] Headplot update the number of channels when loading successive files using playback widget
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/8ac64e18d9206ae28f8aa56ce3a9d12699b865d5

- [x] Add feature to fully address #55 and make a history of playback files show within the playback widget, use JSON to make a history file for 10 entries

- [x] Figure out if 100 history entries can fit in the playback widget using a vertical or horizontal scrollbar

- [x] Dropdown menu to select groups of 10 playback files 
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/b74f99f58c9390cab96fecb93c626c9aa49c4bf3

- [x] add N buttons to main window of playback widget allowing user to load the corresponding playback file, where N is the number of files displayed in the window 1-10

- [x] Update Changelog to reflect changes made in this pull request 
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/457d4a8287a36ec353e2e9105c7d22e8df44c9d1

**Add Feature: Check latest version button** 
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/70d17f0db82ebb0cc8690ea308e89cb91ce96e19 
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/f0f21197b0a9959f873971d1e6b6115a594db1bb
https://github.com/OpenBCI/OpenBCI_GUI/pull/383/commits/4c3cec1181a3bd629bff29a24184a50532ffdb01

- [x] Fetch and parse the latest version from Github using simple code

- [x] create a button

- [x] perform a version check

- [x] allow option to open web browser to latest version page

PR #387, opened 2018-10-09
Fix: Determine number of channels from playback file using thresholds

retiutut:
This is a better method than using switch-case method of counting total columns. Resolves changes for #383


andrewjaykeller:
looks good thanks!

PR #466, opened 2019-03-30
CLP: Dead code

daniellasry:
Found a few instances of what appears to be dead code. Notably:

- InterfaceSerial.pde
- Playground.pde
- Presentation.pde

Would like to check with @conorrussomanno and @retiutut just in case to make sure it's OK to delete.


retiutut:
Presentation mode isn't really "dead". It's included under keyboard shortcuts at the bottom of https://github.com/OpenBCI/Docs/blob/master/OpenBCI%20Software/01-OpenBCI_GUI.md. So if you remove it, you can remove it from the docs, and re-assign the keyboard shortcuts as unused. **Either way, this frees up the Return key to be more useful!**

Need @conorrussomanno's input.

PR #484, opened 2019-04-15
FIX: crash on launch

daniellasry:
The gui was crashing on launch about 1/15 times.

This is fallout from the fix to the gui timing out on launch. I had moved a bunch of the setup() code to a separate thread. As it turns out, ControlP5 does not like being instantiated on a separate thread at the same time as the main gui draw() call is happening. The control panel class instantiates two ControlP5 instances.

Moving the instantiation into the synchronized block will make sure that the control panel instance is not created at the same time as a draw() call is running.


retiutut:
I can confirm this error **was** happening. Cool fix!

PR #496, opened 2019-04-27
Handle exception when loading old settings files

### LINKED ISSUE #493: Catch Error when GUI tries to load old settings files
## Problem

GUI exception happens when trying to load old Settings files. This can happen after new features are added to the GUI that save/load settings.

## Expected 

GUI should handle this automatically in some way and display a gentle message.

macOS/Windows/Linux

## GUI Version

4.1.2

## Running standalone app

Both Standalone and Processing

## Type of OpenBCI Board

All Data Modes

Maybe

retiutut:
Fixes #493

### DIFF
--- a/OpenBCI_GUI/SoftwareSettings.pde
+++ b/OpenBCI_GUI/SoftwareSettings.pde
@@ -39,7 +39,7 @@
 /////////////////////////////////
 class SoftwareSettings {
     //Current version to save to JSON
-    String settingsVersion = "1.0.0"
+    String settingsVersion = "1.0.0";
     //default layout variables
     int currentLayout;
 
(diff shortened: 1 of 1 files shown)


OpenBCI_GUI/SoftwareSettings.pde
```diff
@@ -270,33 +275,18 @@ class SoftwareSettings {
...
-            (!defaultSettingsFileExists) ||
+        if ( //Take a snapshot of the default GUI settings on every system init
             (defaultLoadedDataSource != eegDataSource) ||
             (defaultNumChanLoaded != slnchan)) {
-                println("Saving Default Settings to file!");
+                println("SoftwareSettings: Saving Default Settings to file!");
                 this.save(defaultSettingsFileToSave); //to avoid confusion with save() image
         } else {
-            println("Default Settings file already exists!");
+            println("SoftwareSettings: Default Settings file already exists!");
```
daniellasry: Does this "else" block ever get called, do you need this if/else statement at all?
You are saving the default settings file regardless of wether it already exists or not.

Issue #422, opened 2019-01-13
Analog mode '/2', returns random bad values with 4.0.3. But works with 3.3.1


wjcroft:
## Problem

See this forum post for background. Skiria is using a photoresistor as external trigger on Aux channel. Everything works fine with 3.3.1 standalone. But fails with latest 4.0.3, 4.0.2.

latest comment
http://openbci.com/forum/index.php?p=/discussion/comment/9958/#Comment_9958
on this thread
http://openbci.com/forum/index.php?p=/discussion/1928/timestamps-to-sync-stimulus-and-eeg-vs-external-hardware-trigger

## Expected 

Aux channel should return photoresistor voltage value. Works fine with 3.3.1.

## Operating System and Version

?? macOS/Windows/Linux

## GUI Version

4.0.3 or 4.0.2 standalone fails. But 3.3.1 works.

## Running standalone app

Standalone. But she also tried 4.0.3 with Processing.

## Type of OpenBCI Board

Cyton firmware 3.1

## Are you using a WiFi Shield?

No

girvent:
Windows 8.1 64bit.

brunagirvent:
Windows 8.1 64bit

lopezvaa:
Hi, I had the same problem with the release version 4.1.2, running on windows 10 Enterprise 64 bit.

I ran the GUI code in Processing. There, the digital read functionallity  works fine and the raw file outputs the digital channels as it is meant to. So, the issue is narrowed down to the stand alone release v4.1.2 (beta) that I downloaded while trying to get my wifishield to work, 

Rolling back to release 4.0.4 fixed it (the latest stable release from the Download section is 4.0.03 also works now)

Kind regards
Alejandro

PR #508, opened 2019-05-04
Dev 4.1.2 Minor UI Fixes and Edits

### LINKED ISSUE #418: [USABILITY] Warn users when the dongle and device are not on the same channel
## Problem

It's very confusing for new users if they hit "START SYSTEM" and then "Start Data Stream" but no data is streaming because the dongle and device are not on the same channel (or the device is off!). There is no warning or message to explain why no data is streaming, app behaves as if everything is normal.

When hitting "STATUS", the message "System is down" is pretty vague and does not convey to new users that the device and dongle are not talking at all. Users might assume they are talking, but that something else is wrong with the device.

## Expected 

- When hitting "START SYSTEM" or "Start Data Stream" (ideally both, but at least the latter) warn users that the dongle and the device are not talking (due to channel mismatch, or device being off). A good place for this would be the status bar at the bottom. 

- When hitting "STATUS", we either get "System is up" or "System is down". Replace that with something else that explains that the dongle and the device are either talking or not talking. If they're not talking, suggest the "AUTOSCAN" button.

macOS/Windows/Linux

### LINKED ISSUE #493: Catch Error when GUI tries to load old settings files

### LINKED ISSUE #506: Unstable Digital read on version 4.1.2(beta)
## Problem

I am attempting to set up an external trigger through my cyton board. However, when switching on the digital read via the GUI, I am getting seemingly random numbers from D11, 12, 13, 17,18. (D18 hovers around 8000 and seems to be an accelerometer).  Pushing the prog button should turn D17 from 0 to 1; however, it presents random numbers.

DigitalReadError2

The Raw output files does not record the digital channesl instead, it is as if the accelerometer data was being sampled.

DigitalReadError3

## Expected 

I expected to see the Digital reads as 0s and 1s and see a state change with a TTL level change (or pressing the prog button) and the raw files to contain the 5 columns for the digital inputs rather than the three accelerometer ones.

Windows 10 Enterprise 64 bi

## GUI Version

4.1.2 (beta)

## Running standalone app

Standalone app

## Type of OpenBCI Board

Cyton

No

### LINKED ISSUE #509: Match Time Series drop-down behavior to other graphs
## Problem

The drop-down menus in Time Series widget do not close or allow alternate selection after making a selection. The drop-down must be closed and re-opened to make an additional selection.

Screen Shot 2019-05-05 at 2 48 55 PM

## Expected 

All other drop-downs for the other widgets appear to auto-close after a selection is made, so this is the expected behavior in Time Series, too.

macOS

## GUI Version

4.1.2-beta.0

## Running standalone app

No

## Type of OpenBCI Board

None

No

retiutut:
- [x] help dropdown needs to be shifted over

- [x] UI: Fix Help Button and Settings Button Overlap using Button.ignoreHover

- [x] Fix #418
- [x] Fix #493, but also delete invalid files when detected. Currently, it will be remade when user closes or stops GUI, or saves settings.

- [x] Fix #506 and #422 by setting BoardMode using integers

- [x] Fix #509 

____

- [x] update changelog et al and bump version to 4.1.2-beta.2


OpenBCI_GUI/Radios_Config.pde
```diff
@@ -103,16 +105,19 @@ boolean confirm_connected(){
 /**** Helper function to read from the serial easily ****/
 boolean print_bytes(RadioConfigBox rc){
     if(board_message != null){
-        println(board_message.toString());
-        rc.print_onscreen(board_message.toString());
+        println("Radios_Config: " + board_message.toString());
+        rcStringReceived = board_message.toString().equals("Failure: System is Down") ?
+            "Cyton dongle could not connect to the board. Perhaps they are on different channels? Try pressing AUTOSCAN." :
+            board_message.toString();
```
daniellasry: use a proper if statement here:

```
rcStringReceived = board_message.toString();
if(rcStringReceived.equals("Failure: System is Down")) {
    rcStringReceived += "\nCyton dongle could not connect to the board. Perhaps they are on different channels? Try pressing AUTOSCAN.";
}
```

retiutut: Message is too long

retiutut: Want to expand the message window?

retiutut: Screen Shot 2019-05-04 at 1 40 55 PM

retiutut: Screen Shot 2019-05-04 at 1 42 48 PM

retiutut:
Had to find a different way to check if the Cyton board is connected. Looks like I can check the Cyton firmware version towards the end of initSystem(), then just halt the system if firmwareVersion.equals("v1.0.0").

No one should be using v1.0.0 firmware at this, and this is what appears instead of v3.1.1 when trying to connect to a Cyton board that is powered off.

#418

PR #536, opened 2019-06-03
Relocate SavedData for All OS

### LINKED ISSUE #535: Move All User data to Documents
## Problem

Users for all OS want to find their data easily. On Mac, some users may be restricted from writing files to the Applications folders, due to school/organization policies.

Previous versions of the GUI have always put the data in a "SavedData" folder, but this ends up in the Application directory when using standalone on Mac.

## Expected 

Users can find/create/delete all data generated by the GUI from a location that "makes sense"

OS X, Windows, and possibly Linux

## GUI Version

4.1.2

## Running standalone app

Both

## Type of OpenBCI Board

All boards

Maybe

## Console Log

Most file access shows up in the console log.
Should read:
`OpenBCI_GUI: openNewLogFile: opened ODF output file: /Users/USERNAME/Documents/OpenBCI_GUI/OpenBCI-RAW-2019-06-03_17-46-37.txt`

retiutut:
Fixes #535 

___

- [x] Update GUI doc to reflect this fix (text should be much shorter now)
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ Use OpenBCIHub v2.0.8 please.
 ## Beta 0
 
 ### Bug Fixes
-* GUI now produces valid BDF files #Fixes 441
+* GUI now produces valid BDF files Fixes #441
+* Relocate User data to Documents folder on all OS #535
 
 # v4.1.2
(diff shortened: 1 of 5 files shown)


daniellasry:
For consistency, should we do the same on linux (all 3 OSes)?

PR #537, opened 2019-06-06
Move Sample EEG data to the data folder

retiutut:
Was not included in the 4.1.2 release because it was in the now deprecated /SavedData/ folder.

Proposed fix to make sample data accessible:

- Store sample data in /data/ directory to get distributed on all OS
- check that the files exist in Documents/OpenBCI_GUI/Sample_Data/, if not make directory and copy


retiutut:
This PR only moves the sample data to the /data/ directory.

Issue #59, opened 2016-11-02
Widget: SSVEP


andrewjaykeller:

retiutut:
SSVEP has been merged to development, still needs testing/editing as needed to achieve desired results.

retiutut:
Note: Data processing for SSVEP widget was derived using the following deprecated code - https://github.com/jnaulty/OpenBCI_Processing/blob/ssvep-neurogaming/OpenBCI_GUI/EEG_Processing.pde

PR #569, opened 2019-08-05
Ssvep edits

leanneapichay:
Edited data output to take background frequencies into account when processing


retiutut:
Closes #59

PR #579, opened 2019-08-24
First try at brainflow import

daniellasry:
works fine!
won't merge this PR, just wanna share the code


daniellasry:
Added support fox novaXR, super hacky

PR #594, opened 2019-09-11
Fix for wifi UDPx3 ('burst') mode

bmartin427:
Starting in hub version 1.3.9, requests for the 'udpBurst' protocol from the
GUI were internally remapped to 'udp' protocol with a separate 'burst' field
set to true.  Starting at hub version 2.0.0, it appears this remapping is no
longer done, and instead the GUI command is sent straight to the wifi support
library, which is not intended to recognize 'udpBurst' as a valid protocol.
This protocol name was being passed straight through to a REST URL on the wifi
shield, which did not recognize it and returned a 404 error.  The hub
responded to this 404 error by assuming the wifi shield firmware was out of
date, throwing a RESP_ERROR_WIFI_NEEDS_UPDATE error.

### DIFF
--- a/OpenBCI_GUI/InterfaceHub.pde
+++ b/OpenBCI_GUI/InterfaceHub.pde
@@ -21,6 +21,7 @@ final static String TCP_JSON_KEY_ACTION = "action";
 final static String TCP_JSON_KEY_ACCEL_DATA_COUNTS = "accelDataCounts";
 final static String TCP_JSON_KEY_AUX_DATA = "auxData";
 final static String TCP_JSON_KEY_BOARD_TYPE = "boardType";
+final static String TCP_JSON_KEY_BURST_MODE = "burst";
 final static String TCP_JSON_KEY_CHANNEL_DATA_COUNTS = "channelDataCounts";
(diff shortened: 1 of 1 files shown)


retiutut:
Does this fix a crash scenario?

bmartin427:
Without the fix, the GUI doesn't crash, but the 'UDPx3' protocol option is completely nonfunctional.  It just produces a complaint on the console about the wifi shield firmware needing an update.

bmartin427:
This change to the GUI fixes the issue entirely, without any hub changes.  You could alternatively fix this with exclusively hub changes, without modifying the GUI, but since "udpBurst" isn't really a protocol, it makes sense for the hack pretending that it is to remain confined within the GUI project IMO.

To test, simply test that wifi with UDPx3 works at all, and doesn't just leave you at the Start Session box.  Note that all three protocols for me sometimes take multiple attempts to start, even after this change.

retiutut:
@bmartin427 Can you please do a `git merge upstream/development` into this branch and then I can approve this PR.

retiutut:
@bmartin427 I went ahead and merged this anyways. Please go ahead and do `git merge upstream/development` to get my recent change described in #555

PR #600, opened 2019-09-18
Dev brainflow

daniellasry:
draft PR to see diff

Remove the HUB from OpenBCI_GUI, use brainflow API instead


OpenBCI_GUI/Board.pde
```diff
@@ -0,0 +1,58 @@
...
+
+    public abstract void uninitialize();
+
+    public abstract void update();
+
+    public abstract void startStreaming();
+
+    public abstract void stopStreaming();
+
+    public abstract boolean isConnected();
```
Andrey1994: Can we use BrainFlows's method is_prepared() and remove it here? Or if we really need it in interface I guess it will be better to implement this method like 
```
return boardShim.is_prepared()
```

OpenBCI_GUI/Board.pde
```diff
@@ -0,0 +1,58 @@
...
+    // override this if needed
+    public void setSampleRate(int sampleRate) {
+        outputWarn("Changing the sampling rate is not implemented. Sampling rate will stay at " + getSampleRate());
+    }
+
+    public boolean isAccelerometerActive() { return false; }
+
+    public boolean isAccelerometerAvailable() { return false; }
+
+    public void setAccelerometerActive(boolean active) { }
```
Andrey1994: does it make sense to add warning in these methods?

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -0,0 +1,208 @@
...
+     */
+    abstract protected BrainFlowInputParams getParams();
+    abstract public BoardIds getBoardId();
+
+    protected BoardBrainFlow() {
+        // nothing
+    }
+
+    protected int[] getDataChannels() throws BrainFlowError{
+        return BoardShim.get_eeg_channels(getBoardIdInt());
```
Andrey1994: I think here we should get all channels from 
```
get_eeg_channels()
get_emg_channels()
get_eog_channels()
get_ecg_channels()
```
And maybe even rename this method to something like getEXGChannels

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @Override
+    public boolean initialize() {
+        try {
+            samplingRate = BoardShim.get_sampling_rate(getBoardIdInt());
```
Andrey1994: more likely it will never happen but if there is an exception in samplingRate it will not initialize fields below and looking at exception handler I guess we are able to continue without some data....
I know that it looks bad but lets split it by different try catch blocks(add a function with try catch inside to avoid copypaste?)

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -0,0 +1,208 @@
...
+        
+        if(accelValid) {
+            lastValidAccelValues = lastAccelValues.clone();
+        }
+    }
+
+    @Override
+    public void startStreaming() {
+        println("Brainflow start streaming");
+        if(streaming) {
```
Andrey1994: do we really need to track status here?

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -0,0 +1,208 @@
...
+    private float[] lastValidAccelValues = {};
+
+    /* Abstract Functions.
+     * Implement these in your board.
+     */
+    abstract protected BrainFlowInputParams getParams();
+    abstract public BoardIds getBoardId();
+
+    protected BoardBrainFlow() {
+        // nothing
```
Andrey1994: can we add
```
set_log_file()
enable_board_logger()
```
For BrainFlow logs
Or maybe it can be in static initializer

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -1,3 +1,213 @@
...
+        value = newValue;
+    }
+    public int getValue() { return value; }
+}
+
+static class BoardCytonConstants {
+    static final float series_resistor_ohms = 2200; // Ohms. There is a series resistor on the 32 bit board.
+    static final float ADS1299_Vref = 4.5f;  //reference voltage for ADC in ADS1299.  set by its hardware
+    static final float ADS1299_gain = 24.f;  //assumed gain setting for ADS1299.  set by its Arduino code
+    static final float scale_fac_uVolts_per_count = ADS1299_Vref / ((float)(pow(2, 23)-1)) / ADS1299_gain  * 1000000.f; //ADS1299 datasheet Table 7, confirmed through experiment
```
Andrey1994: why is it here? I do such convertion in BrainFlow, here we just need to calibrate gain. Even if you want to disable convertion in BrainFlow side and restore unscaled data this equation should be reversed

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -1,3 +1,213 @@
...
+    
+    private String port = "";
+    private BoardIds boardId = BoardIds.CYTON_BOARD;
+    private CytonBoardMode currentBoardMode = CytonBoardMode.DEFAULT;
+
+    public BoardCyton(String serialPort, boolean daisy) {
+        super();
+
+        // TODO[brainflow] : WIFI MODE
+        if (daisy) { // 16 channels selected
```
Andrey1994: will it be better to split daisy and plain cyton by different boards?

OpenBCI_GUI/BoardNovaXR.pde
```diff
@@ -0,0 +1,92 @@
...
+    public BoardNovaXR(String ip) {
+        super();
+        ipAddress = ip;
+    }
+
+    // implement mandatory abstract functions
+    @Override
+    protected BrainFlowInputParams getParams() {
+        BrainFlowInputParams params = new BrainFlowInputParams();
+        params.ip_address = ipAddress;
```
Andrey1994: lets get rid of it, ip address can be hardcoded as well as ip protocol

OpenBCI_GUI/BoardNovaXR.pde
```diff
@@ -0,0 +1,92 @@
...
+            println("ERROR: Can't toggle channel " + (channelIndex + 1) + " when there are only " + getNumChannels() + "channels");
+        }
+
+        char[] charsToUse = active ? activateChannelChars : deactivateChannelChars;
+        configBoard(str(charsToUse[channelIndex]));
+    }
+
+    @Override
+    public boolean isAccelerometerActive() {
+        return true;
```
Andrey1994: there are no accel channels

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -0,0 +1,208 @@
...
+    protected int[] accelChannels = {};
+
+    private boolean streaming = false;
+    private float[] lastAccelValues = {};
+    private float[] lastValidAccelValues = {};
+
+    /* Abstract Functions.
+     * Implement these in your board.
+     */
+    abstract protected BrainFlowInputParams getParams();
```
Andrey1994: maybe I would made params and board_id as fields instead methods but its also fine

Andrey1994:
The only thing I dont like - methods like isAnalogAvailable and many more are part of interface, they have default impl and its nice but I dont see any invocations of method isAnalogAvailable for example and dont get why its mandatory.

But if we really need them alternative solution would be to create multiple interfaces/abstract classes with and without these methods and I doubt that it would be better

retiutut:
Adding the JNA package to the libraries folder, this is exactly what fixed LSL streaming.

OpenBCI_GUI/Board.pde
daniellasry: Yes, i think i wrote this before is_prepared was a thing. The BoardBrainflow implementation of isConnected() could return boardShim.is_prepared()

OpenBCI_GUI/Board.pde
daniellasry: Ad in "warning: not implemented on this board"? that kind of thing?

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: I tried that initially, but Contructing the BrainFlowInputParams in the board's constructor if very difficult or inpossible. Because it has to be done in the first line of that board's constructor. Java rule.

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: Yes

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: That's fair

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: Either that, or perhaps we should cancel board initialization and go back to main UI if there is any failire?

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: Maybe i'm being unecessarily careful

OpenBCI_GUI/BoardCyton.pde
daniellasry: You're right. Maybe this is why the data looks like it has the gain applied to it twice

OpenBCI_GUI/BoardCyton.pde
daniellasry: We would save a couple of if statements, but they share a lot of logic.

OpenBCI_GUI/BoardNovaXR.pde
daniellasry: Yep, this was just leftovers from when it wasn't hardcoded yet.

OpenBCI_GUI/BoardNovaXR.pde
daniellasry: This must be why we crash... thanks

daniellasry:
@retiutut I removed JNA because it is already included in the brainflow.jar and I don't want to cause issues with having two versions of the same lib. Are you sure LSL streaming wasn't working? If so, that's weird because I though JNA would be available through brainflow.jar.

OpenBCI_GUI/Board.pde
Andrey1994: yes but I still dont like methods like this in interface but doubt that we can remove them

OpenBCI_GUI/BoardBrainflow.pde
Andrey1994: its not a case right now because all board have package_id but for example if it will be removed later in new boards this packetNumberChannel = BoardShim.get_package_num_channel(getBoardIdInt()); will throw an exception and its fine I guess
Or novaxr has no accel channels and here we get an exception right now I suppose  accelChannels = BoardShim.get_accel_channels(getBoardIdInt()); and its not an error in initialization

OpenBCI_GUI/BoardNovaXR.pde
Andrey1994: yes I also think so

retiutut:
@daniellasry I think you are right, because I tried this briefly last night. I will double check LSL streaming in the standalone GUI.

daniellasry:
> The only thing I dont like - methods like isAnalogAvailable and many more are part of interface, they have default impl and its nice but I dont see any invocations of method isAnalogAvailable for example and dont get why its mandatory.
> 
> But if we really need them alternative solution would be to create multiple interfaces/abstract classes with and without these methods and I doubt that it would be better

I agree that it sucks to have board-specific functionality leaking into the interface. Unfortunately the alternative is worse: we would have to have if-statements all over the gui code to check the board type, then cast the board to the right type. This is not the ideal interface, but it is compatible with the existing GUI code which would be hard to re-write.

The only reason isAnalogAvailable is never called is because I haven't implemented cyton analog mode yet.

I should also mention I've maked some TODO's in the code. Look for `TODO[brainflow]`

Andrey1994:
I think it may be possible to just call get_accel_channels directly in widget for example under try catch block and in this case we dont need to write if statements and dont need to add it to interface

Andrey1994:
But it will not work for non brainflow boards, so lets keep as is

retiutut:
Fixes #669

PR #614, opened 2019-10-01
Option to Draw HeadPlot in a New Window

retiutut:
Objective: Alleviate the stress on the primary application's CPU usage by any means necessary within the limits of Processing/Java.

If we open a widget in a new window in the same fashion as the ConsoleLog, we launch a new "Animation[Animator]" thread in the Java Virtual Machine, well outside of the `OpenBCI.draw()` thread. The effect can be seen in the following GIF:

Screen Recording 2 2019-09-29 at 4 46 01 PM 2019-09-30 23_09_10

After further testing, it seems the new window "ExternalHP" runs at a slower FPS (~22)  while the main window cruises at 59 FPS. I think this will have a positive effect on more crucial functions such as User Interaction and Networking Widget Outputs.

~~Going forward, if this seems like a good idea, we should implement this option for:~~
- [x] ~~TimeSeries~~
- [x] ~~Accelerometer~~
- [x] HeadPlot

Maybe this should be contained to just HeadPlot for now, then SSVEP at a later time. The same pattern will be used.

**Note: This PR is expected to be completed before adding Spectrogram Widget, which is already drawing effectively.**

### DIFF
--- a/OpenBCI_GUI/DataProcessing.pde
+++ b/OpenBCI_GUI/DataProcessing.pde
@@ -4,7 +4,7 @@
 //------------------------------------------------------------------------
 import ddf.minim.analysis.*; //for FFT
 
-DataProcessing dataProcessing;
+static DataProcessing dataProcessing;
 String curTimestamp;
 boolean hasRepeated = false;
 HashMap<Integer,String> index_of_times;
--- a/OpenBCI_GUI/Extras.pde
+++ b/OpenBCI_GUI/Extras.pde
@@ -107,7 +107,7 @@ float calcDotProduct(float[] data1, float[] data2) {
 }
 
 
-float log10(float val) {
+static float log10(float val) {
     return (float)Math.log10(val);
 }
 
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -73,9 +73,9 @@ boolean midInit = false;
 boolean abandonInit = false;
 boolean systemHasHalted = false;
 
-final int NCHAN_CYTON = 8;
-final int NCHAN_CYTON_DAISY = 16;
-final int NCHAN_GANGLION = 4;
+static final int NCHAN_CYTON = 8;
+static final int NCHAN_CYTON_DAISY = 16;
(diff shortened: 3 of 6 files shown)


daniellasry:
Seems pretty useful. Of course it only partially solves the problem since the framerate is still low if you don't move it to a separate window. Can you find a way to do this without duplicating the entire headplot widget?

A better solution would be to move only the expensive headplot calculations to a separate thread. That way those calculations can run at their own pace, in their own update() loop, separate thread from the draw() loop.

Ideally, we would figure out how to cache those calculations or make them more efficient, and if possible avoid the complexity of threads altogether.

By the way, it's not safe to use a boolean as a thread lock. Two threads could try to write to that boolean at the same time (just because it doesn't happen when you test it, doesn't mean it can't happen sometimes, and we will get random crashes here and there). You need to use the proper java Locks and synchronized vars.

retiutut:
I didn't add a thread lock. Originally, it was used only to set the x,y,w,h for the head plot. The main problem I have right now is I have a duplicate `Class HeadPlotE` inside the static function. 
 - Can I copy/clone the `Class HeadPlot` into `static class ExternalHP extends PApplet {`?

I was already planning to do something like this with SSVEP, but I thought I would try it with HeadPlot to see what happened.

retiutut:
> By the way, it's not safe to use a boolean as a thread lock. Two threads could try to write to that boolean at the same time (just because it doesn't happen when you test it, doesn't mean it can't happen sometimes, and we will get random crashes here and there). You need to use the proper java Locks and synchronized vars.

Probably why this happens

```
java.lang.NullPointerException
        at OpenBCI_GUI.doHardCalcs(OpenBCI_GUI.java:20243)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at processing.core.PApplet.method(PApplet.java:3860)
        at processing.core.PApplet$2.run(PApplet.java:3898)
```

Issue #638, opened 2019-11-08
ArrayOutOfBoundsException when streaming 16 channels over network widget


Joe-Westra:
## Problem

When trying to stream data out over the network using the network widget an ArrayOutOfBoundsException is thrown if all of the following conditions hold:
-The number of channels is 16 (either from playback file or from live data)
-The protocol in the network widget is "UDP", "OSC", or "Serial"
-The Data Type is "Time Series"
-"Filters" are turned on

## Expected 

As the unfiltered data stream works fine if "Filters" are set to the off position, I would expect that filtered data would stream out once the Filters are set to On.

## Operating System and Version

Linux Ubuntu 18.04.3

## GUI Version

4.1.6

## Running standalone app

Running from Processing

## Type of OpenBCI Board

Cyton+Daisy
The same issue is experienced when running via the "playback" function with any recorded data.
NOTE: a 16 channel sample included in the /SampleData would be a nice addition!

## Are you using a WiFi Shield?

Yes

## Console Log

I modified the `sendTimeSeriesData() ` method within DataProcessing to print out some extra information to aid in debugging purposes:
```
// TIME SERIES FILTERED
        } else if (filter==1) {
            if (this.protocol.equals("OSC")) {
                for (int i=0;i<nPointsPerUpdate;i++) {
                    msg.clearArguments();
                    for (int j=0;j<numChan;j++) {
                        msg.add(dataBuffY_filtY_uV[j][start+i]);
                    }
                    try {
                        this.osc.send(msg,this.netaddress);
                    } catch (Exception e) {
                        println(e.getMessage());
                    }
                }
            } else if (this.protocol.equals("UDP")) {
                //Added: provide the length of the array
                println("*length of the array: " + dataBuffY_filtY_uV[0].length);

                for (int i=0;i<nPointsPerUpdate;i++) {
                    if (packetNum == MAXPNUM){
                      packetNum = 0;
                    } else {
                      packetNum += 1; 
                    }
                    //Added: packetNumber included in each sample sent
                    String outputter = "{\"type\":\"eeg\",\"packetNumber\":\""+ packetNum +"\",\"data\":[";

                    for (int j = 0; j < numChan; j++) {
                      //Added: Print the current buffer contents
                      println("*buffer data: " + str(dataBuffY_filtY_uV[j][start+i]) + "\tj: " + j + "\tstart+i: " + str(int(start) + int(i)) );
[...]

PR #640, opened 2019-11-08
GUI Networking Bug Fix for 16 Ch

retiutut:
TY @Joe-Westra

Java Exception when streaming 16 channels over network widget w/ high sample rate #638

#Problem#
When trying to stream data out over the network using the network widget an ArrayOutOfBoundsException is thrown if all of the following conditions hold:
-The number of channels is 16 (either from playback file or from live data)
-The protocol in the network widget is "UDP", "OSC", or "Serial"
-The Data Type is "Time Series"
-"Filters" are turned on

_________

- [x] Test bug
- [x] Test fix
- [x] Review Code

### DIFF
--- a/OpenBCI_GUI/W_Networking.pde
+++ b/OpenBCI_GUI/W_Networking.pde
@@ -1096,11 +1096,9 @@ class Stream extends Thread {
 
     Boolean isStreaming;
     Boolean newData = false;
-    // Data buffers
-    int start = dataBuffY_filtY_uV[0].length-11;
-    int end = dataBuffY_filtY_uV[0].length-1;
-    int bufferLen = end-start;
-    float[] dataToSend = new float[numChan*bufferLen];
+    // Data buffers set dynamically in updateNumChan()
+    int start;
+    float[] dataToSend;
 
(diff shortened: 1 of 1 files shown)


retiutut:
@Joe-Westra Can you add "ArrayOutOfBoundsException when streaming 16 channels over network widget #638" as a bug fix in your development branch? It will show up in this PR.

Joe-Westra:
Thanks @retiutut, glad to help out!

OpenBCI_GUI/W_Networking.pde
```diff
@@ -1519,7 +1522,6 @@ class Stream extends Thread {
             } else if (this.protocol.equals("LSL")) {
                 float[] _dataToSend = new float[numChan * 125];
                 for (int i = 0; i < numChan; i++) {
-                    //EEG/FFT readings above 125Hz don't typically travel through the skull
```
retiutut: I just recently added this comment based on discussion with William

retiutut: I will need to add it back @Joe-Westra

PR #678, opened 2020-02-13
minor fixes

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/BoardBrainflow.pde
+++ b/OpenBCI_GUI/BoardBrainflow.pde
@@ -1,4 +1,5 @@
 import brainflow.*;
+import java.util.*;
 import org.apache.commons.lang3.SystemUtils;
 
 abstract class BoardBrainFlow extends Board {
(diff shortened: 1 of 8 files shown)


OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -1,4 +1,5 @@
 import brainflow.*;
+import java.util.*;
```
retiutut: I think this is already declared in OpenBCI_GUI.pde

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -21,19 +22,61 @@ abstract class BoardBrainFlow extends Board {
...
+            println("WARNING: failed to get ecg channels from BoardShim");
+        }
+        try {
+            channels = BoardShim.get_eog_channels(getBoardIdInt());
+            for(int i = 0; i < channels.length; i++) {
+                set.add(channels[i]);
+            }
+        } catch (BrainFlowError e) {
+            println("WARNING: failed to get eog channels from BoardShim");
+        }
```
daniellasry: maybe it's fine to put all of these inside a single try statement. It's unlikely that one will fail but the other will succeed. More likely, they will all fail or all succeed.

PR #679, opened 2020-02-15
Fix DataLogging and General Cleanup - Dev-Brainflow

retiutut:
Part of #600 
- Fixes OpenBCI data format for Cyton w/ Accelerometer
- adds one new method in DataProcessing that works with Brainflow : `saveDataToFile()`
- this method integrates existing OpenBCI data format and BDF format code in DataLogging.pde
- Restore Auto-Connect Cyton button functionality, picks first com port from the list, because it's already reverse sorted, e.g. COM4
- Add a method in Control Panel to find Ganglion BLED112 Dongle Com Port. Designed to work even on Windows for COM1 vs COM2. It does so by looking at the name of the device with jSerialComm Package. You will also find an example in `Networking-Test-Kit/Serial/SerialPortListTest.pde`.

TODO:
- [x] Need help adding a method somewhere to consume `get_timestamp_channel()` from Brainflow boards #680 
- [x] Get timestamp and pass this to `saveDataToFile()` instead of `new Date().getTime()`
- [x] Ensure millisecond accuracy for timestamp saved to file
- [ ] Check that timestamps progress logically (time --->)
- [ ] Add Ganglion record to file also
- [ ] Autoconnect: Handle case when no Dongles are plugged in
- [ ] Test DataLogging on Windows
- [ ] Test on Mac
- [ ] Test on Ubuntu Linux


OpenBCI_GUI/ControlPanel.pde
```diff
@@ -1762,21 +1762,21 @@ class SerialBox {
...
-                _regex = "COM.*$";
-            } else if (isLinux()) {
-                _regex = "^/dev/ttyUSB.*$";
-            }
-            if (ableToConnect(comPort, _regex)) return;
-        } //end for loop for all com ports
+        //Try the last matching comPort. They are already reverse sorted, so get item 0.
+        String comPort = (String)cp5.get(MenuList.class, "serialList").getItem(0).get("headline");
+        if (isMac()) {
+            _regex = "^/dev/tty.usbserial-DM.*$";
```
Andrey1994: I think it will hang again with /dev/ttyxxxx port should be /dev/cuxxxxx

retiutut: I think both of these appear on Mac, and both work. I will look into this again.

Andrey1994: both of them exist but tty hangs while cu doesnt. We tested it with pyserial and brainflow. I have no idea what the difference is.

retiutut:
@Andrey1994 I tried 
```
long timestamp = (long) timestamps[lastReadDataPacketInd % timestamps.length] * 1000;
```

While it is able to be formatted, the result does not have millisecond accuracy:

```
0, -3343.00, -3384.00, -3348.00, -3347.00, -22338.00, -43499.00, -43448.00, -25566.00, 0.168, 0.924, 0.312, 13:20:22.000, 1582140022000
```

Andrey1994:
try 
```
(long) (timestamps[lastReadDataPacketInd % timestamps.length] * 1000);
```
instead
```
(long) timestamps[lastReadDataPacketInd % timestamps.length] * 1000;
```
Also if you keep it as is it will not work with hub boards(synthetic,playback)

Andrey1994:
oh no, these results looks like there is an overflow during multiplication, need to find a way to avoid limit decreasing

retiutut:
gui5 0increased timestamp accuracy

retiutut:
@Andrey1994  When I try the following on Mac
```
long timestamp = (long) (timestamps[pointCounter % timestamps.length] * 1000);
```

 `timestamps.length == 0` and we get division by zero error. 

This does not happen on Windows, and average timestamps size for Cyton 8ch is 5-20. 

**On Mac, it appears as ~120 or 0.**

Andrey1994:
<< On Mac, it appears as ~120 or 0. 
It means buffering of serial port I guess

But 0 is a valid number and in theory it can happen even with buffering disabled, also it means that there are no new data and it should be handled gracefully.

OpenBCI_GUI/DataProcessing.pde
```diff
@@ -60,14 +60,22 @@ int getDataIfAvailable(int pointCounter) {
     if (eegDataSource == DATASOURCE_CYTON) {
```
daniellasry: I assume the code you wrote below will be the same for ganglion, and we won't need this if statement in the future once brainflow ganglion board is implemented?

Looks good for now.

Andrey1994:
lets merge it

retiutut:
I really should have made this Pull Request as a "Draft", so there is still some work to do. Thank you for reviewing existing code!

retiutut:
@daniellasry Ready for review. I can continue work on this in another Draft PR as discussed. Current changes are summarized in the top post on this PR.

OpenBCI_GUI/DataProcessing.pde
```diff
@@ -72,7 +72,10 @@ int getDataIfAvailable(int pointCounter) {
                 //println(timestamps.length);
                 long timestamp = (long) (timestamps[(pointCounter) % (timestamps.length+1)] * 1000);
                 //println(timestamp + " | " + pointCounter % (timestamps.length + 1) + " of " + timestamps.length);
-                saveDataToFile(scaler, lastReadDataPacketInd, timestamp,  currentBoard.getLastValidAccelValues());
+                ///////////////////////////////////////
+                //WORKING HERE TO RESTORE DATALOGGING//
+                ///////////////////////////////////////
+                //saveDataToFile(scaler, lastReadDataPacketInd, timestamp,  currentBoard.getLastValidAccelValues());
```
daniellasry: could you put "//todo[brainflow]" here? i'm using that string to keep track of todos!

PR #682, opened 2020-02-21
wip ganglion brainflow

Andrey1994:
Signed-off-by: Andrey Parfenov <a1994ndrey@gmail.com>

PR #683, opened 2020-02-21
Gui helpers ganglion wip 1

retiutut:
Had to temporarily change hardcoded serial port to COM3 for my windows machine.

- helps with #682 
- removes Bluetooth(Mac built-in) option from the control panel entirely

### DIFF
--- a/OpenBCI_GUI/BoardGanglion.pde
+++ b/OpenBCI_GUI/BoardGanglion.pde
@@ -23,6 +23,7 @@ class BoardGanglion extends BoardBrainFlow {
     
     private String serialPort = "";
     private String macAddress = "";
+    private boolean isCheckingImpedance = false;
 
     public BoardGanglion(String serialPort, String macAddress) {
         super();
@@ -52,5 +53,19 @@ class BoardGanglion extends BoardBrainFlow {
 
         char[] charsToUse = active ? activateChannelChars : deactivateChannelChars;
         configBoard(str(charsToUse[channelIndex]));
+    }
+
+    public void setImpedanceSettings(boolean active) {
+        String s = active ? "Z" : "z";
+        configBoard(s);
+        isCheckingImpedance = active;
+    }
+
+    public void setAccelSettings(boolean active) {
+        configBoard(active ? "N" : "n");
+    }
+
+    public boolean isCheckingImpedance() {
+        return isCheckingImpedance;
     }
 };
--- a/OpenBCI_GUI/BoardGanglion.pde
+++ b/OpenBCI_GUI/BoardGanglion.pde
@@ -56,8 +56,7 @@ class BoardGanglion extends BoardBrainFlow {
     }
 
     public void setImpedanceSettings(boolean active) {
-        String s = active ? "z" : "Z";
-        configBoard(s);
+        configBoard(active ? "z" : "Z");
         isCheckingImpedance = active;
     }
 
--- a/OpenBCI_GUI/BoardGanglion.pde
+++ b/OpenBCI_GUI/BoardGanglion.pde
@@ -56,13 +56,13 @@ class BoardGanglion extends BoardBrainFlow {
     }
 
     public void setImpedanceSettings(boolean active) {
-        String s = active ? "Z" : "z";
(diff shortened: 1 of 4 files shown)


OpenBCI_GUI/BoardGanglion.pde
```diff
@@ -53,4 +54,17 @@ class BoardGanglion extends BoardBrainFlow {
         char[] charsToUse = active ? activateChannelChars : deactivateChannelChars;
         configBoard(str(charsToUse[channelIndex]));
     }
+
+    public void setImpedanceSettings(boolean active) {
```
Andrey1994: I am not sure that its a good idea because there are no such methods in the base class and if we just add them for one specific board we will return to smth like 
```
if cyton
if ganglion
....
```

Andrey1994: but maybe here its ok

retiutut: Setting impedance for Cyton is more complicated, and seems like it is already handled in BoardCyton

```
public void setImpedanceSettings(int channel, char pORn, boolean active) {
        char p = '0';
        char n = '0';

        if (active) {
            if (pORn == 'p') {
                p = '1';
            }
            else if (pORn == 'n') {
                n = '1';
            }
        }

        // for example: z 4 1 0 Z
        String command = String.format("z%c%c%cZ", channelSelectForSettings[channel], p, n);
        configBoard(command);
    }
```

So I copied this method to BoardGanglion, and stripped it down to 1 line, since Ganglion Impedance command is just `z` or `Z`.

retiutut: `isCheckingImpedance()` is used to control when text is drawn to the GanglionImpedance widget.

Andrey1994: in case if this method is common for several boards it should be in the base class or we can add an interface like ImpedanceCompatible and implement this interface for supported boards

retiutut: That sounds very complicated, even though this is a simple solution.

Andrey1994: @daniellasry what do you prefer interface or adding it to the base class? I think even methods like isAnalogActive.... could be available using special interface, but I think its too late to change it

Andrey1994:
I confirm that it works on Windows, and its great!
Lets wait for Daniel to review.

TODO list for ganglion:
* get rid of hardcoded serial port - lets add text input for beginning without autodiscovery
* pass provided serial port to Ganglion board object
* right now device selection field is just a mock - device which you specify is not selected in fact

Mostly all changes are related to control panel.

OpenBCI_GUI/BoardGanglion.pde
Andrey1994: lets keep as is

daniellasry: Rest of the change looks good. I think just adding the ImpedanceCompatible  interface would be nice. ALternatively I can just send a PR to your PR so you see how i did it since it's a small change

Andrey1994: << if (currentBoard instanceof BoardCyton || currentBoard instanceof BoardGanglion || currentBoard instanceof BoardNovaXR)
I think in practice it will be even worse because we will need to cast currentBoard object to specific type so more likely it will be like
if cyton
if ganglion

<< Instead we just check if they are an instance of ImpedanceCompatible. The idea is that the GUI remains agnostic of specific board implementations.
Exactly and in a perfect world with perfect GUI this approach should be more widely used I think

retiutut: @daniellasry Can you make a PR to address this? You and Andrey seem to agree on updating the approach to impedance, and I think y'all know what's best.

retiutut:
@Andrey1994 @daniellasry I'm merging this request as is, since this was just a break-fix PR.

OpenBCI_GUI/BoardGanglion.pde
daniellasry: Yes, working on it

Issue #648, opened 2019-11-14
Feature: Cyton Impedance Check Headplot


retiutut:
**Is your feature request related to a problem? Please describe.**
Takes too long to check impedance on all channels when using Cyton and Cyton+Daisy.

**Describe the solution you'd like**
A way to test impedance on all Cyton channels at once, and possibly display this on a headplot diagram.

**Describe alternatives you've considered**
Current solution requires users to click all 16buttons on and off. Could also implement an Expert Mode keyboard shortcut.

**Additional context**
https://openbci.com/forum/index.php?p=/discussion/2364/feature-request-impedance-check-map#latest

![f1hiqcU](https://user-images.githubusercontent.com/25914123/68816045-dc309100-0642-11ea-83e8-e8862c1cf5e2.png)

markopetrovic-neurotech:
Cant this be done simply by swapping the eeg data used in Headplot with the global impedence data variable?

  headPlot.setIntensityData_byRef(dataProcessing.data_std_uV, is_railed);
->         headPlot.setIntensityData_byRef(data_elec_imp_ohm, is_railed);

retiutut:
@markopetrovic-neurotech It's technically correct but not so simple. If you look at the image above, the issue arises when users choose something other than the default electrode positions. We need to expand the electrode map to something like what's shown above and make it configurable/assignable to channels.

markopetrovic-neurotech:
Thanks for the timely responses!

I thought we could load other electrode positions from electrode_position.txt files, or is the issue for mappings with more than 16 electrodes? 

Also I see the position electrode_xy array is being updated when the electrodes are dragged, it seems to be working but I'm not sure if I can tell the difference.

Thanks again, I'm excited for the upcoming milestone and would love to help where I can.

retiutut:
> Thanks for the timely responses!
> 
> I thought we could load other electrode positions from electrode_position.txt files, or is the issue for mappings with more than 16 electrodes?
> 
> Also I see the position electrode_xy array is being updated when the electrodes are dragged, it seems to be working but I'm not sure if I can tell the difference.
> 
> Thanks again, I'm excited for the upcoming milestone and would love to help where I can.

Woah... I just learned something new about the GUI! But I think that moving them to an arbitrary XY position is a "hacky" solution in comparison to showing a full 10-20 diagram and clicking the ones we want.


Very interesting.... thanks for sharing again!

markopetrovic-neurotech:
[...]

PR #685, opened 2020-02-27
Made an interface defining functions for impedance settings boards.

daniellasry:
BoardCyton and BoardNovaXR implement interface `ImpedanceSettingsBoard` since they both share the `setImpedanceSettings()` function.

In Ganglion, I renamed `setImpedanceSettings` to `setCheckingImpedance`. Is had the same name as the Cyton/NovaXR function but it did something completely different... I did not make an interface for `setCheckingImpedance` and `isCheckingImpedance` because Ganglion is the only one that needs those.

### DIFF
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -23,7 +23,7 @@ static class BoardCytonConstants {
     static final float leadOffDrive_amps = 6.0e-9;  //6 nA, set by its Arduino code
 }
 
-class BoardCyton extends BoardBrainFlow {
+class BoardCyton extends BoardBrainFlow implements ImpedanceSettingsBoard {
     private final char[] deactivateChannelChars = {'1', '2', '3', '4', '5', '6', '7', '8', 'q', 'w', 'e', 'r', 't', 'y', 'u', 'i'};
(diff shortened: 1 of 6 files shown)


daniellasry:
I am going to do the same thing with Accel, Analog, Digital, and Marker. Make interfaces for them and take them out of Board.pde

Andrey1994:
<< I am going to do the same thing with Accel, Analog, Digital, and Marker. Make interfaces for them and take them out of Board.pde

Thats great! But I am afraid that it will be much bigger change

Andrey1994:
Stupid question but I see widget for impedance checking using Ganglion only(no idea what exactly it does but doesnt matter right now) but I dont see widgets for checking impedance for Cyton/Novaxr...

daniellasry:
From what i understand, 'Impedance check" is a ganglion-only thing. But i don't understand what it does. I think Richard/Conor are the only ones who do

retiutut:
@Andrey1994 @daniellasry Impedance check for Cyton boards is baked into Hardware Settings Controller in TimeSeries. There is already a customer request and issue made to make Cyton impedance check more like GanglionImpedance widget. https://github.com/OpenBCI/OpenBCI_GUI/issues/648

PR #688, opened 2020-03-05
Remove board-specific stuff from softwaresettings

daniellasry:
- Removed channel settings and other board-specific settings out of software settings.
- Make a helper function for changing the framerate, minimize duplicate code

### DIFF
--- a/OpenBCI_GUI/Interactivity.pde
+++ b/OpenBCI_GUI/Interactivity.pde
@@ -879,21 +879,22 @@ void toggleFrameRate(){
         frameRateCounter = 1; // until we resolve the latency issue with 24hz, only allow 30hz minimum (aka frameRateCounter = 1)
     }
     if(frameRateCounter==0){
-        frameRate(24); //refresh rate ... this will slow automatically, if your processor can't handle the specified rate
-        topNav.fpsButton.setString("24 fps");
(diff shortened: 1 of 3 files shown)


retiutut:
This completely removes the code that was used to set hardware settings. I thought we would just move it for now. Not sure I can approve this right now. Can you please move this code or comment it out?

retiutut:
I am aware that this change does speed up Session Start, but I would still like to have the code until Brainflow integration is done.

daniellasry:
I don't think this code should have been here in the first place, and I don't think it should ever come back. This is for your own good. I can bring it back, and ultimately you'll have to live with maintaining it long term :)

Happy to discuss more at the next meeting though. There's tons of stuff in software settings that really irks me...

daniellasry:
And remember... it's all saved in github. I fyou want to bring code back you just revert the commit. No need to leave a whole bunch of commented out dead code everywhere. That's another thing I don't like about our code base. Big blocks of commented out code everywhere.

retiutut:
@daniellasry Thank you for explaining to me. I will accept this change, since you feel strongly that it should not be there.

Andrey1994:
<<  That's another thing I don't like about our code base. Big blocks of commented out code everywhere.
Completely agreed, its terrible and its not only in GUI, its everywhere and I was really surprised when I saw it for the first time

PR #689, opened 2020-03-05
Board interfaces - Remove board-specific fucntionality from board interface

daniellasry:
Made 4 interfaces:

- DigitalCapableBoard
- AnalogCapableBoard
- AccelerometerCapableBoard
- MarkerCapableBoard

Boards that have the feature inherit from the interface.

Cleaned up some duplicate code along the way

### DIFF
--- a/OpenBCI_GUI/AnalogCapableBoard.pde
+++ b/OpenBCI_GUI/AnalogCapableBoard.pde
@@ -1,0 +1,7 @@
+
+interface AnalogCapableBoard {
+
+    public boolean isAnalogActive();
+
+    public void setAnalogActive(boolean active);
+};
--- a/OpenBCI_GUI/BoardBrainflow.pde
+++ b/OpenBCI_GUI/BoardBrainflow.pde
@@ -11,9 +11,9 @@ abstract class BoardBrainFlow extends Board {
     protected int[] dataChannels = {};
     protected int[] accelChannels = {};
 
-    private boolean streaming = false;
-    private float[] lastAccelValues = {};
-    private float[] lastValidAccelValues = {};
+    protected boolean streaming = false;
+    protected float[] lastAccelValues = {};
+    protected float[] lastValidAccelValues = {};
 
     /* Abstract Functions.
      * Implement these in your board.
@@ -235,11 +235,6 @@ abstract class BoardBrainFlow extends Board {
     @Override
     public int getNumChannels() {
         return dataChannels.length;
-    }
-
-    @Override
-    public float[] getLastValidAccelValues() {
-        return lastValidAccelValues;
     }
 
     public int getBoardIdInt() {
--- a/OpenBCI_GUI/BoardBrainflow.pde
+++ b/OpenBCI_GUI/BoardBrainflow.pde
@@ -2,7 +2,7 @@ import brainflow.*;
 import java.util.*;
 import org.apache.commons.lang3.SystemUtils;
 
-abstract class BoardBrainFlow extends Board {
+abstract class BoardBrainFlow implements Board {
     private DataPacket_ADS1299 dataPacket;
     private BoardShim boardShim = null;
 
--- a/OpenBCI_GUI/BoardBrainflow.pde
+++ b/OpenBCI_GUI/BoardBrainflow.pde
@@ -251,6 +251,11 @@ abstract class BoardBrainFlow extends Board {
         configBoard(command);
     }
 
+    @Override
+    public void setSampleRate(int sampleRate) {
+        outputWarn("Changing the sampling rate is not possible on brainflow boards. Sampling rate will stay at " + getSampleRate());
+    }
+
     protected void configBoard(String configStr) {
         if(!isConnected()) {
             outputError("Cannot send " + configStr + " to board. The board is not connected");
--- a/OpenBCI_GUI/W_Accelerometer.pde
+++ b/OpenBCI_GUI/W_Accelerometer.pde
@@ -57,6 +57,8 @@ class W_Accelerometer extends Widget {
     boolean accelInitHasOccured = false;
     private Button accelModeButton;
 
+    AccelerometerCapableBoard accelBoard;
+
     W_Accelerometer(PApplet _parent) {
         super(_parent); //calls the parent CONSTRUCTOR method of Widget (DON'T REMOVE)
 
@@ -88,6 +90,9 @@ class W_Accelerometer extends Widget {
         accelModeButton.textColorNotActive = color(255);
         accelModeButton.hasStroke(false);
         accelModeButton.setHelpText("Click to activate/deactivate the accelerometer!");
+
+        // This widget is only instantiated when the board is accel capable, so we don't need to check
+        accelBoard = (AccelerometerCapableBoard)currentBoard;
     }
 
     void initAccelData() {
@@ -133,7 +138,7 @@ class W_Accelerometer extends Widget {
     void update() {
         super.update(); //calls the parent update() method of Widget (DON'T REMOVE)
 
-        if (isRunning && currentBoard.isAccelerometerActive()) {
+        if (isRunning && accelBoard.isAccelerometerActive()) {
             //update the current Accelerometer values
             updateAccelPoints();
             //update the line graph and corresponding gplot points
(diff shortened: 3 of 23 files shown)


OpenBCI_GUI/AnalogCapableBoard.pde
```diff
@@ -0,0 +1,7 @@
+
+interface AnalogCapableBoard {
```
Andrey1994: I've tried to implement analog/digital mode for cyton and realized that I completely misunderstand how to return data and what kind of data.... For some reason its not like for accel where we have 
```
public float[] getLastValidAccelValues();
```
It will be good to add it to interface but since I really dont get existing pipeline for such data I know better

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -11,9 +11,9 @@ abstract class BoardBrainFlow extends Board {
     protected int[] dataChannels = {};
     protected int[] accelChannels = {};
 
-    private boolean streaming = false;
-    private float[] lastAccelValues = {};
-    private float[] lastValidAccelValues = {};
+    protected boolean streaming = false;
+    protected float[] lastAccelValues = {};
+    protected float[] lastValidAccelValues = {};
```
Andrey1994: Should it implement AccelerometerCapableBoard interface if it has these fields?

OpenBCI_GUI/W_Accelerometer.pde
```diff
@@ -88,6 +90,9 @@ class W_Accelerometer extends Widget {
         accelModeButton.textColorNotActive = color(255);
         accelModeButton.hasStroke(false);
         accelModeButton.setHelpText("Click to activate/deactivate the accelerometer!");
+
+        // This widget is only instantiated when the board is accel capable, so we don't need to check
+        accelBoard = (AccelerometerCapableBoard)currentBoard;
```
Andrey1994: Should here be a check like
```
currentBoard instanceof AccelerometerCapableBoard
```

OpenBCI_GUI/W_AnalogRead.pde
```diff
@@ -94,6 +96,8 @@ class W_AnalogRead extends Widget {
         } else {
             analogModeButton.setHelpText("Click this button to activate/deactivate analog read on Cyton pins A5(D11), A6(D12) and A7(D13).");
         }
+
+        analogBoard = (AnalogCapableBoard)currentBoard;
```
Andrey1994: And the same check here and in all other widgets

OpenBCI_GUI/W_Accelerometer.pde
Andrey1994: Ah, I see its done in widgetmanager

Andrey1994:
@daniellasry I think its great change!

OpenBCI_GUI/W_Networking.pde
```diff
@@ -1253,33 +1253,7 @@ class Stream extends Thread {
...
-                                    sendAccelerometerData();
-                                }
-                            } else if (this.dataType.equals("Focus")) {
-                                sendFocusData();
-                            } else if (this.dataType.equals("Pulse")) {
-                                sendPulseData();
-                            } else if (this.dataType.equals("SSVEP")) {
-                                sendSSVEPData();
-                            }
+                            sendData();
```
retiutut: Awesome!

OpenBCI_GUI/W_Networking.pde
```diff
@@ -1373,6 +1321,43 @@ class Stream extends Thread {
         }
     }
 
+    void sendData() {
```
retiutut: I like this!

OpenBCI_GUI/WidgetManager.pde
```diff
@@ -47,13 +47,15 @@ void setupWidgets(PApplet _this, ArrayList<Widget> w){
...
+    if (currentBoard instanceof AccelerometerCapableBoard) {
+        //Widget_2
+        w_accelerometer = new W_Accelerometer(_this);
+        w_accelerometer.setTitle("Accelerometer");
+        addWidget(w_accelerometer, w);
+    }
 
     //only instantiate this widget if you are using a Ganglion board for live streaming
-    if(nchan == 4 && eegDataSource == DATASOURCE_GANGLION){
+    if(nchan == 4 && currentBoard instanceof BoardGanglion){
```
retiutut: This new format for checking board type reads better than before.

OpenBCI_GUI/WidgetManager.pde
```diff
@@ -105,29 +107,34 @@ void setupWidgets(PApplet _this, ArrayList<Widget> w){
...
-    if(eegDataSource == DATASOURCE_CYTON){
+    if(currentBoard instanceof BoardCyton){
         //Cyton Widget_8
         w_pulsesensor = new W_PulseSensor(_this);
         w_pulsesensor.setTitle("Pulse Sensor");
         addWidget(w_pulsesensor, w);
         // println("  setupWidgets pulse sensor -- " + millis());
+    }
 
+    if(currentBoard instanceof DigitalCapableBoard) {
```
retiutut: 👍

OpenBCI_GUI/AnalogCapableBoard.pde
daniellasry: It;s a mess! The way it works: the DataPacket_ADS1299 class has "auxvalues" and these can be digital, analog, or accel. all 3. It's gross I would love to refactor it soon.

Accelerometer is the exception, where getLastValidAccelValues() is the official way to get accel values. I would like to make analog and digital work similarly, or refactor it all to be completely different...

retiutut: It wasn't easy to make it work, and there is/was weird code on converting these incoming "auxvalues".

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: Potentially... not all brainflow boards are accel capable. Maybe I have to rethink this.

OpenBCI_GUI/W_AnalogRead.pde
daniellasry: Maybe i should add the same comment as above here

OpenBCI_GUI/BoardBrainflow.pde
Andrey1994: There is a simple solution but it will limit us to use only brainflow supported devices....
We can store only entire data array returned by BrainFlow which contains all kinds of data and split it only in widgets using methods like
```
get_eeg_channels()
```
Also it will require refactoring in files like DataProcessing.pde and I really dont want to touch them

OpenBCI_GUI/AnalogCapableBoard.pde
daniellasry: We'll make it better when we tackle analog and digital

OpenBCI_GUI/BoardBrainflow.pde
Andrey1994: And I dont fully understand why we need these variables here since AccelCapableBoard interface has a method to request data, if we need temp buffers to store this data it can be defined only in boards which implement this interface

daniellasry: I'll take a second look and audit this stuff to make sure it's necessary.

Agreed that anything accel-speific should be moved to the boards that need it. Like you said, BoardBrainflow can store the entire data array and then BoardCyton can extract the aux channels out of it.

OpenBCI_GUI/WidgetManager.pde
daniellasry: Should be safer too, because you know for sure the board is comaptible before proceeding.

Hopefully we can phase out the usage of  "eegDataSource" eventually. It should just be a private variable in the control panel to indicate which board is selected. We'll use it to instantiate the right board type nothing else.

OpenBCI_GUI/BoardBrainflow.pde
daniellasry: What really needs to happen here is a refactoring of how data flows through the GUI. That's a big enough change to be its own PR so i'm gonna merge this and work on a separate fix.

PR #694, opened 2020-03-11
Cleanup Cyton AutoConnect button backend

retiutut:
- [x] Check for the first available comPort with name "FT231X USB UART" on all OS
- [x] @daniellasry "...attempting to start the session with every port is a bad idea anyway. Instead, we should be using the same logic as the "Status" button. Check the status of each port until one returns "success"." <--- This would also handle situation when more than one Cyton dongle is plugged into the same computer, as they should already be setup on different radio channels. Only the correct Cyton would respond with "success" to paired dongle.
- [x] Change the name of `board` to `serial_direct_board` for specific use in control panel and radios_config.pde
- [x] Debug the 'Status' button


daniellasry:
I think there's a bug (at least on windows) where the "status" button always return "success". I wonder if it's related to this. 

I think it happened because I deleted or disconnected some code related to serial stuff. I have yet to track down the change where it broke.

retiutut:
I vaguely remember trying this at one point. It's possible that the SerialPort will be busy when trying to connect via Brainflow for Start Session. 

I can look into this again / try the above action item, now that the backend has changed.

retiutut:
>I think there's a bug (at least on windows) where the "status" button always return "success". I wonder if it's related to this.

#698

OpenBCI_GUI/Radios_Config.pde
```diff
@@ -121,12 +121,34 @@ boolean print_bytes(RadioConfigBox rc){
...
+    }
+}
+
 void print_bytes_error(RadioConfigBox rcConfig){
     println("Radios_Config: Error reading from Serial/COM port");
     rcConfig.print_onscreen("Error reading from Serial port.\n\nTry a different port?");
     board = null;
 }
 
+void print_bytes_error(){
```
retiutut: All of the changes in this file are duplicate methods that exclude printing to the RadioConfig textBox. Should I just combine them and use `if (rcConfig != null) `?

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -1751,38 +1751,32 @@ class SerialBox {
...
-            String[] foundCytonPort = match(_comPort, _regex);
-            if (foundCytonPort != null) {  // If not null, then a match was found
-                println("ControlPanel: Connect using comPort: " + _comPort);
-                openBCI_portName = foundCytonPort[0];
+        String comPort = getCytonComPort();
+        if (comPort != null) {
+            openBCI_portName = comPort;
+            if (system_status()) {
+                board.stop(); //Stop serial port connection
+                board = null;
```
Andrey1994: nitpicking but lets call board.uninitialize() manually if possible, just board = null is ok cause it will trigger destructor which calls release_session in brainflow and it will release all resources.
But first of all maybe one day or for one board or for non-brainflow boards it will not call release_session/uninitialize and another possible issue - destructor will be called by garbage collector and in theory we can not ensure that garbage collector will be destroy it before we decide to do smth else with this serial port

retiutut:
Cyton Radio Config Status button passes the "button mashing" test, for me at least.

PR #701, opened 2020-03-17
add accel for ganglion

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -74,12 +74,10 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         configBoard(str(charsToUse[channelIndex]));
     }
 
-    @Override
     public boolean isAccelerometerActive() {
         return getBoardMode() == CytonBoardMode.DEFAULT;
     }
 
-    @Override
     public void setAccelerometerActive(boolean active) {
         if(active) {
             setBoardMode(CytonBoardMode.DEFAULT);
@@ -87,7 +85,6 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         // no way of turning off accel.
     }
 
-    @Override
     public float[] getLastValidAccelValues() {
         return lastValidAccelValues;
     }
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -74,10 +74,12 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         configBoard(str(charsToUse[channelIndex]));
     }
 
+    @Override
     public boolean isAccelerometerActive() {
         return getBoardMode() == CytonBoardMode.DEFAULT;
     }
 
+    @Override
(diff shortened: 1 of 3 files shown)


OpenBCI_GUI/BoardCyton.pde
```diff
@@ -74,20 +74,17 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         configBoard(str(charsToUse[channelIndex]));
     }
 
-    @Override
```
daniellasry: why remove the @ override?

OpenBCI_GUI/BoardGanglion.pde
```diff
@@ -37,8 +38,28 @@ class BoardGanglion extends BoardBrainFlow {
         configBoard(str(charsToUse[channelIndex]));
     }
 
-    public void setAccelSettings(boolean active) {
+    @Override
+    public boolean initialize()
+    {
+        // turn on accel by default, or is it handled somewhere else?
```
daniellasry: This is fine. It's not on by default in firmware?

OpenBCI_GUI/BoardCyton.pde
Andrey1994: because it doesnt override)
override was required when we had only Board class which contained all methods. Now its in interface and there is no default implementation. Interface methods should not be marked by override

Andrey1994: but maybe its fine to mark interface methods using override keyword, I dont remember such details of java code style. At least both options work

retiutut: http://www.javapractices.com/topic/TopicAction.do?Id=223

retiutut: Eager to learn how the above link applies to this situation.

OpenBCI_GUI/BoardGanglion.pde
Andrey1994: I am not sure but I think its disabled and for sanity check can keep it here or somewhere else in the code

OpenBCI_GUI/BoardCyton.pde
Andrey1994: According to this article since jdk7 its nice to mark interface methods by override keyword but its not required. Before jdk7 it was only to override super class methods.

Personally I dont like the idea to use override for interface methods because intuitively it makes me think that there is super class with such method and maybe I need to write smth like:
```
super.some_method()
// my code
```
And it will not work obviously because for interface there is no such method in super class...

But I dont want to argue with java code style, will return @ override

daniellasry: Good point, with an interface it's not really needed because you'll get a compiler warning if you don't implement it correctly. (not the case if you are overriding a fully defined function in a base class)

daniellasry: @ override was never required, but it would give you a compiler error if you didn't override it properly (typo in the function name, different arguments, or if the base class function changed). 

But yeah, not required for interface functions because the compiler will force you to implement those correctly.

Andrey1994: I think for interface it will be an error not warning if you dont implement it correctly. But not sure

daniellasry: You are correct

PR #706, opened 2020-03-28
Make PlaybackBoard for Playback Mode

retiutut:


retiutut:
Compiles, but the data in TimeSeries is choppy.

OpenBCI_GUI/BoardPlayback.pde
```diff
@@ -0,0 +1,131 @@
...
+        streaming = false;
+    }
+
+    public boolean isConnected() {
+        return isInitialized;
+    }
+
+    @Override
+    public int getSampleRate() {
+        return 250;
```
daniellasry: It can't possibly be 250 always. It depends on the sampling rate of the board when it was recording.

retiutut: Yes, this information is in the header of the CSV files.

PR #711, opened 2020-04-06
Data Processing Refactor

daniellasry:
- [ ] Refactor data processing flow 
- [ ] Add analog, digital, marker
- [ ] Make it easy to add other aux data types


daniellasry:
Recalling this PR since I think the datapacket class shoudl go away

PR #714, opened 2020-04-14
Make timeseries faster

daniellasry:
Drawing GPlot lines in batches for better performance. See GLayer.java
Updating plot points in separate threads per channel.

Observed about a 60% performance improvement (18fps to 29fps on my machine).

This could potentially be sped up more by having each channel render to a different PGraphics object on a different thread, and have the main thread combine the results.


Andrey1994:
Why do we merge it to development branch instread dev-brainflow?

daniellasry:
Because it's not brainflow or novaXR specific. We can merge it up into that branch

PR #717, opened 2020-04-15
stop streaming for hardware settings

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/W_TimeSeries.pde
+++ b/OpenBCI_GUI/W_TimeSeries.pde
@@ -271,10 +271,16 @@ class W_timeSeries extends Widget {
                 if(showHardwareSettings) {
                     showHardwareSettings = false;
                     hsc.isVisible = false;
+                    if (isRunning) {
+                        currentBoard.startStreaming();
+                    }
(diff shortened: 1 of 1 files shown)


Andrey1994:
> The "Stop Data Stream" button needs to change text and color to ~ "Start Data Stream". I can look at this later.
> 
> aka front-end stuff.

I think its not required, mostly because as soon as you close hardware settings window streaming will be restarted automatically, so you dont need to click at this button

retiutut:
That's a break in data, and it affects playback files also! I really need to look at this first before merging.

retiutut:
It needs to hard stop the data stream, and cause users to click "Start Data Stream".

Andrey1994:
Users dont need to click start stream because stream will be started as soon as settings are closed

retiutut:
Solution for Later: 
- Auto-restart stream like Andrey suggests
- Grey out and disable "Start Stream button" while hardware settings controller is open

Andrey1994:
there is one more option, if streaming is running and somebody clicks on hardware setting button instead openning settings widget create a message like 'you need to stop streaming before changing settings'

PR #723, opened 2020-04-22
Novaxr aux

Andrey1994:


OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -15,6 +15,7 @@ abstract class BoardBrainFlow implements Board {
     protected boolean streaming = false;
     protected float[] lastAccelValues = {};
     protected float[] lastValidAccelValues = {};
+    protected double[][] rawData = null;
```
daniellasry, line 18: I learned that in java, you can have an array of length zero. SO you can initialize like this:
```protected double[][] rawData = new double[0][0];```

and you won't have to check if rawData is null later on.

Comment (Andrey1994) [reply OpenBCI_GUI/BoardBrainflow.pde:18]: To get size of nested array I will still need to write smth like
```
rawData[0].length
```
And it will fail

daniellasry: ah, fair enough

PR #726, opened 2020-04-23
wip novaxr aux but works for window size 20

Andrey1994:

Create a PR to show the diff.

image

Its WIP and I think we should replace FixedStack class by smth more powerful and use it in all widgets but for now its ok. Ideally we should port this https://github.com/Andrey1994/fps_inspector_sdk/blob/master/src/Utils/inc/DataBuffer.h to java(or find smth similar and thread safe)
Its the same ringbuffer which I use inside brainflow but in brainflow its optimized for double arrays and uses my own spinlocks while here its template based and uses windows specific locks.

@retiutut it should help you to continue. Also I wrote some todos and comments please take a look and fix.

@daniellasry would be nice if you take a look also

It works but only for window size == 20. I create stack with size ==  nPoints which is 20 by default and use it in timey calculations and if you set window size < 20 it doesnt work...

### DIFF
--- a/OpenBCI_GUI/W_NovaAux.pde
+++ b/OpenBCI_GUI/W_NovaAux.pde
@@ -1,8 +1,7 @@
+import java.util.*; 
+
 
 class W_NovaAux extends Widget {
-
-    //to see all core variables/methods of the Widget class, refer to Widget.pde
-    //put your custom variables here...
 
     private int numAnalogReadBars;
     float xF, yF, wF, hF;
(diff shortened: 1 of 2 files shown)


Andrey1994:
And nice thing - if we use this approach we dont need all this horrifying code with indexes...

Andrey1994:
@retiutut I think eda and ppg have completely different orders of values, it needs to be visualizes correctly

OpenBCI_GUI/W_NovaAux.pde
```diff
@@ -342,57 +304,26 @@ class AuxReadBar{
...
-
-                GPoint tempPoint = new GPoint(timey, voltage);
-                auxReadPoints.set(i-buffDiff, tempPoint);
-
-            }
-            plot.setPoints(auxReadPoints); //reset the plot with updated auxReadPoints
+        int id = 0;
+        while (enu.hasMoreElements()) { // there are exactly nPoints elements
+            float timey = -(float)numSeconds + (float)id * timeBetweenPoints;
+            System.out.println("time " + timey);
```
retiutut, line 318: Just need to remove this println() used for debugging

retiutut:
@Andrey1994 Here is the fix for stopping data while not running:
```
// update data in plot
        if (isRunning) {
            updatePlotPoints();
        }
```

Andrey1994:
> @Andrey1994 Here is the fix for stopping data while not running:
> 
> ```
> // update data in plot
>         if (isRunning) {
>             updatePlotPoints();
>         }
> ```

nice, feel free to fix it directly in this branch

Andrey1994:
> Looks good to me. The FixedStack thing is achieving the same thing as the "AppendAndShift" function, but obviously this is much better.

I think in data pipeline refactoring we need to implement advanced FixedStack

retiutut:
I like the idea of using existing Java functions over "duct-tape solution" like putting AppendAndShift() everywhere.

PR #731, opened 2020-04-23
Data Flow Refactor Round 1

daniellasry:
``Board`` is now an abstract base class that hold a FixedStack buffer and accumulates data from the board subclasses into it.

It provides two public methords for data retrieval:

```
    // returns the data collected on the current frame
    public double[][] getFrameData();
    // return a sublist of the FixedStack
    public List<double[]> getData(int maxSamples);
```

Board subclasses must implement the following protected methods:
```
    // used internally by the Board base class to accumulate data in the FixedStack
    protected abstract double[][] getNewDataInternal();
    protected abstract boolean initializeInternal();
    protected abstract void uninitializeInternal();
    protected abstract void updateInternal();
    protected abstract int getTotalChannelCount();
```

The Board base class will call getNewDataInternal() on every update and accumulate the results in the FixedStack.

Datatype-specific board interfaces will expose the channel indices for that data type:
```
interface AccelerometerCapableBoard {
    public boolean isAccelerometerActive();
    public void setAccelerometerActive(boolean active);
    public int[] getAccelerometerChannels();
};
```

Widgets can get data directly from the board object using getData(int). Then they can narrow down to the data type they need using getAccelerometerChannels(), getAnalogChannels(), etc.

- Updated all current boards to use this new data flow
- Deleted uncecessary code in DataProcesing and W_Accelerometer
- Cleaned up unnecessary variables and buffers (many more left)
- Updated TimeSeries, Accelerometer, and NovaAux to get data from board using new method


retiutut:
This is cool! Much better to have clear get() methods for data.

daniellasry:
In this PR as it stand right now, I attempted to use board_shim.get_current_board_data() and never flush the internal buffer. Then I pass the data directly to the widgets for drawing.

This works, exept for a few concerns:
1. When starting and stopping the stream, the plot is cleared and starts over. This is because when you stop and start stream, it clears the internal board_shim buffer. That's not expected behavior for the gui.
1. The networking widget sends data on every frame, so we **need** to know what data was received from the board this frame. There's no way to know with get_current_board_data().
1. For long recordings of many hours, the internal board_shim buffer could get too big. And if we only write playback file at the end, if the app crashes we lose all data.
1. Small performance hit as several widgets are calling board_shim.get_current_board_data() on every frame. 

I am going to change it so that it each board calls get_board_data() on every update and stores the result internally. This addresses all the points above:

1. Widgets will ask the board for data on this frame, and accumulate it in their own FixedStack.
1. The networking will ask the board for the current frame data and send it.
1. We are clearing the internal buffer and writing playback file data every frame.
1. Loss performance hit as we only call get_board_data() once per frame and store the result.

Andrey1994:
```
// returns all the data this board has received in this frame
public double[][] getDataThisFrame();
```
So we will continue wo unified data collection interface?

Andrey1994:
And I dont like existing interface of FixedStack, for example it doesnt allow to request specified amount of data and it may be a problem if we change window size....(thats why eda\ppg widget works only for window size 20). Also its not thread safe and we can push/pop only one element at time, its not effective.... I wrote it as a simple temp solution to complete eda\ppg widget

daniellasry:
I didn't see a need for a unified data collection interface. In a sense the board is the unified interface. I want to hear your thoughts if you think we need one though.

We can make FixedStack better in the future. Right now there is no need for thread safety and eda/ppg widget works fine for other window sizes in this PR.

Andrey1994:
I mean right now we need to create FixedStack inside every widget a little better option would be to change board interface from
```
public double[][] getDataThisFrame();
```
to 
```
public double[][] getLastData(int dataCount) //(from multiple frames) in BrainFlow such method may return less data if there are not enought data points
```
Implementation of this method may use BrainFlow's method or advanced FixedStack. 

If we do this widgets will not need to create instances of FixedStack and will be able to request data from multiple frames

Andrey1994:
Or we can implement both methods in board interface
```
get_current_data(int num_samples)
get_frame_data()
```

daniellasry:
That's exactly how I had it in revision c3d4d32. I am open to going back to it.

I had: ```public double[][] getData(int dataCount)```

Each widget would call getData() on the board with however many samples the widget needs. Then the widget would assign that data directly into the GPlot. 

I wrote about the issues I had:
1. When starting and stopping the stream, the plot is cleared and starts over. This is because when you stop and start stream, it clears the internal board_shim buffer. That's not expected behavior for the gui.
1. The networking widget sends data on every frame, so we **need** to know what data was received from the board this frame. There's no way to know with get_current_board_data().
1. For long recordings of many hours, the internal board_shim buffer could get too big. And if we only write playback file at the end, if the app crashes we lose all data.
1. Small performance hit as several widgets are calling board_shim.get_current_board_data() on every frame. 

We could address 1. if the board_shim did not clear its internal buffer when stopping the stream. I'm not sure how we would fix 2, 3, 4.

daniellasry:
I guess we could fix 2. If we implemented both methods like you said. Then the board would have to keep track of many samples it received this frame. On every update() call, it would have to compare the board_data_count() with last update call.

Andrey1994:
"When starting and stopping the stream, the plot is cleared and starts over. This is because when you stop and start stream, it clears the internal board_shim buffer. That's not expected behavior for the gui."
Actually stop_stream doesnt flush buffer, start_stream creates a new one. This issue can be solved if we create FixedStack buffer on GUI side(only one).

"For long recordings of many hours, the internal board_shim buffer could get too big"
Not at all, its ring buffer with fixed max size.

"Small performance hit as several widgets are calling board_shim.get_current_board_data() on every frame."
There is a lock but it does memcpy and its fast enough

daniellasry:
Yes, we could create a single FixedStack in the gui to accumulate all data. But it would have to be one FixedStack per channel, correct? Doing this would also solve the performance problem, since we're only getting data from the board_shim once (instead of once per widget).

Ah ok, so if the internal buffer has a max size, we need to be writing to the playback file regularly. We can't wait until the end of the session to write.

get_current_board_data() has a performance hit because it allocates/destroys/copys new arrays with every call, and also does reshape_data> It's fast, but if every widget is calling it on every frame it becomes noticeable.

Andrey1994:
"Yes, we could create a single FixedStack in the gui to accumulate all data. But it would have to be one FixedStack per channel, correct?"
No In BrainFlow I store arrays in ringbuffer we can do the same

Andrey1994:
"get_current_board_data() has a performance hit because it allocates/destroys/copys new arrays with every call, and also does reshape_data. It's fast, but if every widget is calling it on every frame it becomes noticeable."
It will be done only once in Board class, we can call get_current_board_data in update method, store result in a variable and read from this variable in widgets

Andrey1994:
"Ah ok, so if the internal buffer has a max size, we need to be writing to the playback file regularly. We can't wait until the end of the session to write."
I am not sure but I think existing gui also asks for session duration to allocate buffer with fixed size for playback... There is "Max File Duration" input in UI

Andrey1994:
So I think it may look like this:

* Board calls BrainFlow's method get_board_data() and stores it in GUI's advanced FixedStack as arrays with only one FixedStack for all channels. BrainFlow's reshape method will be executed only once per update here
* There is a method in Board Interface to request specified amount of data(or less if there are less data? or maybe preinit it with 0?) from this single FixedStack which should return double[][] or ArrayList<double[]>
* Widgets call this method together with get_eeg_channels and process data
* For networking or maybe some other widgets we can intoduce second method to Board Interface like get_frame_data

daniellasry:
Okaya, we can make a single FixedStack<Double[]> with arrays.
How big should this FixedStack be? Should it be the maximum window length of 20seconds * sampleRate?

Every time a widget asks for the data, do we need to convert from FixedStack<Double[]> to double[][]? maybe not, we can just return the fixed stack? If we can return the FixedStack, we would need to improve it so widgets can ask for specific amount of data, as you said.

If we implement it this way we should be able to keep track of the data received this frame (it's whatever data was added to the fixedStack this frame).

Andrey1994:
"Should it be the maximum window length of 20seconds * sampleRate?"
I would add ~20% to it for safety if sampling rate is not accurate

"do we need to convert from FixedStack<Double[]> to double[][]?"
I think so

"we would need to improve it so widgets can ask for specific amount of data, as you said."
Yep

"If we implement it this way we should be able to keep track of the data received this frame (it's whatever data was added to the fixedStack this frame)."
yes, just one more method

BTW I really dont like networking widget, its inaccurate to stream data only on update calls. If we stream on the SDK level we would send data with constant speed instead chunks at 30 FPS(and in code it would look much better)

Andrey1994:
But if its harder to implement lets continue with existing code as in PR, its also good and definitely better than before

Andrey1994:
""do we need to convert from FixedStack<Double[]> to double[][]?"
I think so"
But we dont need to convert entire fixedstack to double[][], just requested amount of data

daniellasry:
Thanks Andrey, this is helpful.
I'm trying different things and I want to find the best solution. I like the idea of a single FixedStack because it ensures that all widgets behave the same. I will attempt that implementation.

I was incorrect about the networking widget. Actually it waits until it has 40ms worth of data, and send that chunk. It sends data in 40ms chunks.

daniellasry:
But yeah I don't like the Networking widget either. The code is a mess (big surprise) and I bet most of its features aren't being used that much.

Andrey1994:
"The code is a mess (big surprise) and I bet most of its features aren't being used that much." 

Agreed, and I dont like current approach "put everything to gui" it leads to huge amount of features that are not used and code is messy. And we can not deprecate/remove them. IMO some widgets must be implemented as separated applications outside gui, like SSVEP, Networking, etc. But unfortunately we continue putting everything inside gui

daniellasry:
Totally, the GUI is a result of adding more and more features that maybe 1 person asked for. WIthout any consideration of cost, maintainability... It's trying to be a research tool but it's not. And unfortunately we're still adding stuff (#713)

daniellasry:
Made some changes, let me know what you think. 

- Board interface is now an abstract base class. The FixedStack is stored in the Board base class.
- On update, Board class will call getNewDataInternal() (abstract function) and append the data to the FixedStack<double[]>
- Public interface for retrieving data is:

```
    public double[][] getFrameData();
    public List<double[]> getData(int maxSamples);
```

These two functions are implemented in the Board base class. getData() returns a subset of the FixedStack by using the subList() function.

getFrameData() is not used right now but will be used by data logging code.

daniellasry:
I don't want this change to get too big so I'm opening it up for review. I will continue with architectural improvements in other PRs.

A few things on the to-do list:
- Write class that collect data from board and logs to playback file (I believe it is faster to reqrite datalogging code than to try and fix current code).
- Write Board class that reads from playback file.
- Fix up the Networking widget (doesn't look to difficult)
- Hook up Cyton analog data to analog widget
- Add Cyton digital data to Brainflow and hook it up to the GUI
- Add Cyton Marker mode to the GUI
- Reorganize DataProcessing class
  - Rename buffers used by DataProcessing to be more descriptive
  - Move buffers inside the DataProcessing class instead of global
  - Delete unnecessary buffers
- Audit entire code, and move any global variables/functions appropriately
- Clean up all Hub-related code

Andrey1994:
"Add Cyton digital data to Brainflow and hook it up to the GUI"
There is such data in brainflow, it returns raw bytes in get_other_channels and I dont want to introduce new methods

Andrey1994:
I think you can make better decision here but maybe in widgets it would be nice to have method like
```
double[][] get_current_data(int size, double defaul_value)
```
instead 
```
double[][] get_current_data(int maxSize)
```
Mostly it depends on how we plot data in widgets.
Idea - we may need to plot data for window size == 5 and there are less data. If we use method like ```
double[][] get_current_data(int maxSize)
```
we need to calc position in widget code and I think add default value for missed data(0 most of the time)

**Comment by daniellasry** [comment]
> "Add Cyton digital data to Brainflow and hook it up to the GUI"
> There is such data in brainflow, it returns raw bytes in get_other_channels and I dont want to introduce new methods

Alright, should be fine to get digital/marker from there if needed.

``double[][] get_current_data(int size, double defaul_value)``

I haven't need to do this because the FixedStack is filled with zeroes by default and all the widgets (so far) use zero as their default value.

Also the FixedStack is sized large enough to accomodate 23 seconds of data, which is more than then maximum of 20 all widgets use as their windows. No object should be asking for more than 20s of data. If they do, then the FixedStack will be made bigger.

**Comment by Andrey1994** [comment]
"I haven't had to do this because the FixedStack is filled with zeroes by default and all the widgets (so far) use zero as their default value."
Yep I did it temporary. Do you think its a good idea?

File: OpenBCI_GUI/OpenBCI_GUI.pde [hunk 1]
```diff
@@ -1037,33 +1031,10 @@ void systemUpdate() { // for updating data values and variables
...
-            }
-
-            //New feature to address #461, defined in DataLogging.pde
-            //Applied to OpenBCI Data Format for LIVE mode recordings (Cyton and Ganglion)
-            //Don't check duration if user has selected "No Limit"
-            //TODO[brainflow] commented this out to get brainflow working. Undo comment and fix.
-            // if (outputDataSource == OUTPUT_SOURCE_ODF
-            //     && eegDataSource < DATASOURCE_PLAYBACKFILE
-            //     && settings.limitOBCILogFileDuration()) {
-            //         fileoutput_odf.limitRecordingFileDuration();
```
retiutut, line 1061: I've added a note on Clubhouse to repair this feature after data refactor.

daniellasry: thanks!

daniellasry:
> "I haven't had to do this because the FixedStack is filled with zeroes by default and all the widgets (so far) use zero as their default value."
> Yep I did it temporary. Do you think its a good idea?


the fill() function? yeah, i think it's convenient.

PR #735, opened 2020-04-27
Fix writing of GUI playback files and simplify code

daniellasry:
- Added abstract function so that Board implementations can provide names for their data columns
- Improved openBCI data format to have names over each column (file can be opened as .csv)
- Created DirectoryManager class
  - Currently only provides date string for use in files names (replacing getDateString())
  - In the future, I want this class to manage and generate GUI directories (settings directory, session directory, console data, etc.)
  - This functionality is currently in SoftwareSettings but it doesn't belong there
- Turned DataLogger.pde into a class
  - added update(), initialize(), uninitialize(), onStartStream() and onStopStream() to activate the data logger.
  - moved all the logic to start/stop recording into this class. this is now triggered by the functions above.
- Largely simplified DataWriterODF, simply writes raw data from board to file.
- Thanks to uninform board interface, any board can have logging, not just Cyton and Ganglion
- Move isChannelActive() to the board interface.

Coming in a later change:
- Generate session names for NovaXR board too
- Fix BDF data logging too
- Minor refactorings
  - Get rid of outputDataSource
  - Move stuff like `settings.isLogFileOpen()` and  `settings.setLogFileStartTime` into DataLogger class, it does not belong in settings.


daniellasry:
I realized i need to think of a way to format the files so that when reading them back, we know what each column is for.

Should we read the column names to determine what they are? Not ideal because the column names coudl be any string.

Maybe we should add channel mapping infor the the file header? something like:

```
%packetnumchannel=0
%exgchannels=1, 2, 3, 4, 5, 6, 7, 8
%accelchannels=9, 10, 11
...
```

retiutut:
Yes we need to improve the header in CSV output. It would also be great to specify Analog, Digital, or Accel data. 

This would allow Cyton users to playback Aux data using the GUI. Before, just looks weird in the accel widget.

Andrey1994:
> I realized i need to think of a way to format the files so that when reading them back, we know what each column is for.
> 
> Should we read the column names to determine what they are? Not ideal because the column names could be any string. Those are just there to make the data pretty and human-readable.
> 
> Maybe we should add channel mapping infor the the file header? something like:
> 
> ```
> %packetnumchannel=0
> %exgchannels=1, 2, 3, 4, 5, 6, 7, 8
> %accelchannels=9, 10, 11
> ...
> ```

Maybe its not an option for GUI but in BrainFlow I thought about it either and realized that I dont need it. Data returned by read_file method has exactly the same format as data returned by board.get_board_data() and it means that all functions like BoardShim.get_eeg_channels(id) work fine for this data.

daniellasry:
Andrey, that's a good idea. I don't need to store the channel numbers in the file, only the board type. Then I can ask the appropriate board class for the channel numbers.

daniellasry:
> Yes we need to improve the header in CSV output. It would also be great to specify Analog(3 values), Digital(5 values), or Accel(3 values) data.
> 
> This would allow Cyton users to playback Aux data using the GUI. Before, just looks weird in the accel widget, with no way to differentiate.

This is already implemented in this code. Currently only accel channels will be labeled. Analog and digital are showing up under "other". But once Digital and Analog are fuly implemented they will automatically show up properly labeled.

retiutut:
@daniellasry Awesome, that will be much nicer than before. Let me try this branch out right now!

retiutut:
I like that the human readable timestamp in the last column now includes the date Important for overnight recordings! 

> Currently only accel channels will be labeled. Analog and digital are showing up under "other". But once Digital and Analog are fuly implemented they will automatically show up properly labeled.

Lot's of "other" right now lol. But if it will be labeled, that will make it possible to improve GUI Playback Mode.

```
%OpenBCI Raw EEG Data
%Number of channels = 4
%Sample Rate = 1600 Hz
Sample Index, EXG Channel 0, EXG Channel 1, EXG Channel 2, EXG Channel 3, Accel Channel 0, Accel Channel 1, Accel Channel 2, Other, Other, Other, Other, Other, Other, Other, Other, Other, Other, Timestamp, Timestamp (Formatted)
0.0, 15.075536727905273, 16.148887634277344, 13.08965015411377, 26.147510528564453, 2.688000202178955, -1.8560000658035278, -31.584001541137695, 192.0, 0.0, 84.0, 255.0, 198.0, 252.0, 37.0, 0.0, 0.0, 0.0, 1.588038018091302E9, 2020-04-27 20:40:18.091
```

daniellasry:
Yep, lots of "other". It's easier in code to just give users all the data than to try and exclude columns. And besides, I don't think anyone will complain about getting all of the data.

PR #736, opened 2020-04-28
Made it so uninitialized boards can still return channel indices.

daniellasry:
My idea for BoardPlayback is to have it create a dummy instance of the original board internally, and use that instance to get channel indices.

For example, if we open a Cyton playback file, BoardPlayback will create a dummy instance of BoardCyton and use it to call getEXGChannels(), getAccelerometerChannels(), etc.

This is the first step towards that goal. Those functions need to work without having to call initalize() on the board.

### DIFF
--- a/OpenBCI_GUI/BoardBrainflow.pde
+++ b/OpenBCI_GUI/BoardBrainflow.pde
@@ -5,11 +5,12 @@ import org.apache.commons.lang3.SystemUtils;
 abstract class BoardBrainFlow extends Board {
     private BoardShim boardShim = null;
 
-    protected int samplingRate = 0;
-    protected int packetNumberChannel = 0;
-    protected int timeStampChannel = 0;
-    protected int totalChannels = 0;
-    protected int[] exgChannels = {};
+    protected int samplingRateCache = -1;
+    protected int packetNumberChannelCache = -1;
+    protected int timeStampChannelCache = -1;
+    protected int totalChannelsCache = -1;
+    protected int[] exgChannelsCache = null;
+
     protected boolean streaming = false;
 
     /* Abstract Functions.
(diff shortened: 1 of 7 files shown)


OpenBCI_GUI/DataWriterODF.pde
```diff
@@ -22,6 +22,7 @@ public class DataWriterODF {
         output.println("%OpenBCI Raw EEG Data");
         output.println("%Number of channels = " + nchan);
         output.println("%Sample Rate = " + currentBoard.getSampleRate() + " Hz");
+        output.println("%Board = " + currentBoard.getClass().getName());
```
retiutut, line 25: Sweet!

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -5,11 +5,12 @@ import org.apache.commons.lang3.SystemUtils;
...
-    protected int samplingRate = 0;
-    protected int packetNumberChannel = 0;
-    protected int timeStampChannel = 0;
-    protected int totalChannels = 0;
-    protected int[] exgChannels = {};
+    protected int samplingRateCache = -1;
+    protected int packetNumberChannelCache = -1;
+    protected int timeStampChannelCache = -1;
+    protected int totalChannelsCache = -1;
+    protected int[] exgChannelsCache = null;
```
retiutut, line 12: >This is the first step towards that goal. Those functions need to work without having to call initalize() on the board.

is that what these `...cache` variables help with?

daniellasry: Right. After this PR, if you make a new instance of a board, you don't have to call instantiate() before you can call getExtChannels(), getAccelChannels(), etc.

However, everytime you call one of those functions, it asks the boardShim for the data. This does a lot of work, parsing a json file... To avoid doing this work more than once, I cache the result in those variables after the first call.

PR #738, opened 2020-04-29
Remove W_OpenBionics

daniellasry:
Looks like dead code, usused for a long time.


retiutut:
Correct, this is old code. I think this was used to simulate controlling a 3D printed hand.

Not really something **everyone** can use.

PR #739, opened 2020-04-29
Read and display playback files

daniellasry:
- #### Playback board is not a streaming board. So it doesn't need the FixedStack<>. The FixedStack<> would make it difficult to do scrubbing in playback.
  - Created an interface `DataSource`. This defines the public interface for all objects that provide data.
  - `Board` implements `DataSource`
  - All physical boards (cyton, ganglion) inherit from `Board`
  - `DataSourcePlayback` implements `DataSource`.

- #### Implemented `DataSourcePlayback`.
  - Implements all `*Capable` interfaces
  - Parses header to determine the board that generated the data
  - Creates an instance of this underlying board internally. Uses underlying board for channel indices and other info.
  - Reads all data into an ArrayList<double[]>
  - Keeps track of current sample index, update() function increments this index when streaming (based on sample rate and time elapsed since last update()).
  - getData() returns data array using ArrayList.subList() (avoids copy)
  - If more data is requested than available, pad with zeroes

- #### For this to work, all boards need to have a default contructor, and there needs to be a 1-1 mapping between board class and Brainflow board ID.
  - I created helper classes `BoardCytonSerial, BoardCytonSerialDaisy, BoardCytonWifi, BoardCytonWifiDaisy`which inherit from the abstract base class `BoardCyton`
  - Same for Ganglion.

- #### Fixed up timeseries for playback.
  - Improved time formatting
  - Allow live scrubbing even with stream stopped

- Fixed a bug where playback recent file dropdown was not displaying the right file names.
- When changing playback file mid-session, it will restart the session instead of trying to fix up arrays... (reinitRequested)
- Deleted stale code


daniellasry:
> Just one question: we use float buffers instead double buffers... is it because graphic library we use doesnt support double? Can we change it to double to avoid casting everything to float?

Yes, that's the reason. I was hoping to avoid it too but it's [not possible](https://github.com/jagracar/grafica/blob/master/src/grafica/GPoint.java). 

We have the source code for the gfx library, and we could modify it. But eventually it uses java/processing draw calls which take floats. Which itself is using GL to draw lines. So i think the cast will have to happen somewhere.

retiutut:
I'm taking some time to absorb these changes. I want test this locally and see the moving parts.

OpenBCI_GUI/W_Accelerometer.pde
```diff
@@ -122,7 +122,7 @@ class W_Accelerometer extends Widget {
     void update() {
         super.update(); //calls the parent update() method of Widget (DON'T REMOVE)
 
-        if (visible && updating && isRunning && accelBoard.isAccelerometerActive()) {
+        if (visible && updating && accelBoard.isAccelerometerActive()) {
```
retiutut, line 125: This probably allows the accelerometer data to display when scrubbing the playback file.

OpenBCI_GUI/W_Accelerometer.pde
```diff
@@ -483,13 +482,4 @@ class AccelerometerBar {
         plot.setDim(w - 36 - 4 - xOffset, h);
 
     }
-
-    void clearAllGPlots() {
```
retiutut, line 487: This was the fix before improved playback scrubbing in this PR.

OpenBCI_GUI/W_Playback.pde
```diff
@@ -199,46 +199,23 @@ void playbackSelectedWidgetButton(File selection) {
     } else {
         println("W_Playback: playbackSelected: User selected " + selection.getAbsolutePath());
         playbackFileSelected(selection.getAbsolutePath(), selection.getName());
-        if (playbackFileIsEmpty) {
-            haltLoadingFile(selection.getAbsolutePath());
-        } else {
-            reInitAfterPlaybackSelected();
-        }
+        reInitAfterPlaybackSelected();
```
retiutut, line 202: You got it!

OpenBCI_GUI/W_Accelerometer.pde
daniellasry: You are correct

OpenBCI_GUI/W_Accelerometer.pde
daniellasry: Right, no longer needed!

OpenBCI_GUI/W_Playback.pde
```diff
@@ -271,30 +244,18 @@ void playbackFileSelected (String longName, int listItem) {
...
-    } else {
-        outputError("playbackFileSelected: Data table appears to be null! Please submit an issue on GitHub!");
-        return;
-    }
+
     //Output new playback settings to GUI as success
     outputSuccess("You have selected \""
-    + shortName + "\" for playback. "
-    + str(nchan) + " channels found.");
+    + longName + "\" for playback.");
```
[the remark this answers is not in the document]
daniellasry: Thanks, used that for debugging and forgot to revert.

OpenBCI_GUI/W_Spectrogram.pde
```diff
@@ -464,21 +464,15 @@ class W_Spectrogram extends Widget {
...
-                    return index_of_times.get(currentTableRowIndex);
-                }
-            } else {
-                //This is a sanity check for null exception, and this would print on screen
-                return "TimeNotFound";
-            }
-        } else {
-            //Same here
-            return "TimeNotFound";
+        if (index_of_times != null && index_of_times.get(0) != null) { //Check if the hashmap is null to prevent exception
```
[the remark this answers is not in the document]
daniellasry: Yep, I've added a todo and opened a ticket in clubhouse

PR #740, opened 2020-04-29
fix sampling rate issue

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -93,12 +93,8 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
         boolean res = super.initializeInternal();
         if ((res) && (samplingRateCache > 0)){
-            try {
-                String command = samplingRateCommands.get(samplingRateCache);
-                boardShim.config_board(command);
-            } catch (Exception e) {
-                e.printStackTrace();
-            }
+            String command = samplingRateCommands.get(samplingRateCache);
+            sendCommand(command);
         }
         return res;
     }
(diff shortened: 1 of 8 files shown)


OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -686,7 +686,7 @@ void initSystem() throws Exception {
             // TODO[brainflow] : do we need these two lines?
             int nEEDataValuesPerPacket = nchan;
             boolean useAux = true;
-            currentBoard = new BoardCyton(openBCI_portName, wifi_ipAddress, nchan == 16, selectedProtocol == BoardProtocol.WIFI);
+            currentBoard = new BoardCyton(openBCI_portName, wifi_ipAddress, nchan == 16, selectedProtocol == BoardProtocol.WIFI, selectedSamplingRate);
```
retiutut, line 689: Makes sense when looking at this line of code!

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -79,7 +91,16 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
...
 
-        return super.initializeInternal();
+        boolean res = super.initializeInternal();
+        if ((res) && (samplingRateCache > 0)){
+            try {
+                String command = samplingRateCommands.get(samplingRateCache);
+                boardShim.config_board(command);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
```
daniellasry, line 101: You can use sendCommand() here instead, it already does the error checking. Same for ganglion.

Andrey1994: I thought in java's hashmap get method throws exception but it returns null instead, so yep you are right

PR #743, opened 2020-04-29
Fix cyton impedance

daniellasry:
It was off by 1


retiutut:
As long as it works and you've tested it!

PR #744, opened 2020-04-30
Fix digital read for cyton

daniellasry:
Thankfully this wasn't difficult at all. Tested it by connecting pins to gound.

### DIFF
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -169,6 +169,9 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
     @Override
     public int[] getDigitalChannels() {
+        // the removeAll function will remove array indices 0 and 5.
+        // remove other_channel[0] because it's the end byte
+        // remove other_channels[5] because it does not contain digital data
         int[] digitalChannels = ArrayUtils.removeAll(getOtherChannels(), 0, 5); // remove non-digital channels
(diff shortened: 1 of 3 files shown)


OpenBCI_GUI/BoardCyton.pde
```diff
@@ -169,9 +169,8 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
     @Override
     public int[] getDigitalChannels() {
-        // TODO[brainflow]
-        return new int[0];
-        // return digitalChannels;
+        int[] digitalChannels = ArrayUtils.removeAll(getOtherChannels(), 0, 5); // remove non-digital channels
```
Andrey1994, line 172: I dont understand it. Format for other_channels in Cyton is: cyton end byte, 6 bytes of raw data. What should this line do?

retiutut:
It's happening using Processing 3 and 4. :/

OpenBCI_GUI/BoardCyton.pde
daniellasry: I'll add a comment to explain better.
There are only 5 digital pins, and 7 channels in other_channels.
I remove 0 because it's the end byte
I remove 5 because it does not contain digital data
And only return the relevant channels

retiutut:
Tested this just now using same method: Ground->Pin.

Some of them just appear to be on by default, but when you plug in a pin and send UP or DOWN correctly, the indicator will be accurate.

PR #746, opened 2020-05-02
Add travis and appveyor auto-deployment 

daniellasry:
- Modified appveyor.yml
  - run the make-release script with no prompts
  - upload fully built .zip to s3 bucket openbci-gui
  - uploads to `<branchname>/<commit-id>` as well as `<branchname>/latest`
  - also pushed it as artifact

- Created .travis.yml
  - linux installation is simple
    - download and unzip processing
    - add processing-java to path
    - copy libraries

  - mac installation
    - download and unzip processing
    - use the processing-java bash script checked-in with this PR (this script is created by processing IDE on mac)
    - move Processing.app to /Applications/. Bash script requires it
    - chmod +x processing-java
    - add processing-java to path
    - copy libraries
    - install dmgbuild

  - run the make-release python script with no prompts

- Removed Hub stuff from make-release.py
- Removed 32-bit windows release
- Specify output path for processing-java (fails without it in CI)
- Verified make-release.py still works when run manually on local machine

We can download and share built files using [this link](https://demo.filestash.app/s/L4vY5Lt)
Which was set up using a read-only access key

Missing for now:
  - Windows signing
  - Mac signing


daniellasry:
> Cool! It doesnt send linux/macos builds to AWS.. Are there any issues with it?

It actually does! Travis now supports publishing to s3 via the 'artifacts' add-on. The bucket and keys are set as environment vars in the travis project settings.

Appveyor also supports pushing artifacts to s3 directly, but only to one folder. I wanted to copy to two folder (commit-id and "latest") so I had to do with with awscli there.

retiutut:
GUI 4.2.0 was built using Processing 4, can we start using that? Upstream issue involving serial ports has been fixed.

retiutut:
>Removed 32-bit windows release

Thanks!

retiutut:
Will the builds be available on the Github releases page? This is were www.openbci.com/downloads fetches links from.

daniellasry:
Is it safe to use processing 4? I see that it's in alpha. Do we know how long until it's a stable release?

This PR will not auto-publish anything to the releases page yet. It also won't sign the app (yet).
Before we can auto-publish to the releases page I'll have to take care of signing.

For the time ebing we can access the file using [the link](https://demo.filestash.app/files/?share=L4vY5Lt). (we can share this link with novaxr folks for example)

retiutut:
@daniellasry It is safe to use Processing 4, as I have used it for months. There was only one issue I found, and a fix has been merged. Still, Alpha2 release has not been made yet, which would include the fix. I just asked when they pan to make the next release.

>This PR will not auto-publish anything to the releases page yet. It also won't sign the app (yet).
Before we can auto-publish to the releases page I'll have to take care of signing.

>For the time ebing we can access the file using the link. (we can share this link with novaxr folks for example)

Good things. If we can publish to releases page in the future using same vX.X.X format, the openbci.com/downloads page will continue to work just fine.

release_script/mac_only/about_processing-java.txt
```diff
@@ -0,0 +1 @@
+On mac, the only way to install processing-java is to open the Processing IDE and go to Tools->Install "processing-java". We can't do this in travis. Since processing-java is just a simple bash script that runs some jar files, we can copy it here and install the latest verison of processing separately during the travis run.
```
retiutut, line 1: Thanks for this note!

release_script/make-release.py
```diff
@@ -144,47 +132,6 @@ def package_app(sketch_dir, flavor, windows_signing=False, windows_pfx_path = ''
     else:
         print ("Successfully deleted source dir.")
 
-    ### Ask user for the hub directory
-    ###########################################################
-    hub_dir = input("Drag and drop the HUB for " + flavor + " [ENTER to skip]: ")
```
retiutut, line 149: yes!

PR #747, opened 2020-05-03
Repair W_AnalogRead after Data Refactor

retiutut:
Screen Shot 2020-05-02 at 8 54 24 PM

### DIFF
--- a/OpenBCI_GUI/W_AnalogRead.pde
+++ b/OpenBCI_GUI/W_AnalogRead.pde
@@ -261,6 +261,7 @@ class AnalogReadBar{
     int lastProcessedDataPacketInd = 0;
 
     int[] analogReadData;
+    private AnalogCapableBoard analogBoard;
 
     AnalogReadBar(PApplet _parent, int _analogInputPin, int _x, int _y, int _w, int _h) { // channel number, x/y location, height, width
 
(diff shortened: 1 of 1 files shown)


OpenBCI_GUI/W_AnalogRead.pde
```diff
@@ -338,23 +339,41 @@ class AnalogReadBar{
...
+    void initArrays() {
+        nPoints = nPointsBasedOnDataSource();
+        timeBetweenPoints = (float)numSeconds / (float)nPoints;
+        analogReadPoints = new GPointsArray(nPoints);
+
+        for (int i = 0; i < nPoints; i++) {
+            float time = calcTimeAxis(i);
+            float analog_value = 0.0; //0.0 for all points to start
+            analogReadPoints.set(i, time, analog_value, "");
+        }
```
daniellasry, line 354: Now that you have the initArray() function you need to delete lines 311 to 325 (inclusive) and replace them with a single call to initArrays().

daniellasry: Also, delete the "arBuffSize" class variable since it's not used anymore.

daniellasry: Also, delete the "analogReadData" calss variable.

daniellasry: While you're there, please make all class variables "private". (or as many of them as possible right now)

OpenBCI_GUI/W_AnalogRead.pde
```diff
@@ -338,23 +339,41 @@ class AnalogReadBar{
...
     void update() {
 
-        //update the voltage value text string
-        float val;
-
-        //update the voltage values
-        val = hub.validAccelValues[auxValuesPosition];
-        analogValue.string = String.format(getFmt(val),val);
+         // early out if unactive
+        if (auxValuesPosition == 1) {
```
daniellasry, line 362: Why do you need this if statement? If auxValuesPosition is something other than 1 it should not early out?

OpenBCI_GUI/W_AnalogRead.pde
```diff
@@ -369,48 +388,23 @@ class AnalogReadBar{
...
+        return -(float)numSeconds + (float)sampleIndex * timeBetweenPoints;
+    }
 
-            int voltage = dataPacketBuff[lastProcessedDataPacketInd].auxValues[auxValuesPosition];
+    void updatePlotPoints() {
+        List<double[]> allData = currentBoard.getData(nPoints);
+        int[] channels;
 
-            analogReadData[analogReadData.length - numSamplesToProcess + samplesProcessed] = voltage; //<>//
+        channels = analogBoard.getAnalogChannels();
```
daniellasry, line 399: no need to declare and set channels onseparate lines

retiutut:
AnalogRead Networking Data type will also need to be repaired.

Andrey1994:
The only thing I dont like(and its almost everywhere in our code) is casting without type check....

daniellasry:
> The only thing I dont like(and its almost everywhere in our code) is casting without type check....

Are you referring to `analogBoard = (AnalogCapableBoard) currentBoard;` ?
The widget is only created if the board is accelerometer capable, so we should never arrive here without the right type of board.

daniellasry:
> > The only thing I dont like(and its almost everywhere in our code) is casting without type check....
> 
> Are you referring to `analogBoard = (AnalogCapableBoard) currentBoard;` ?
> The widget is only created if the board is accelerometer capable, so we should never arrive here without the right type of board.
> 
> Anywhere else, we should do a type check first.

Andrey and I talked offline and we agree this is fine because we do the type check in WidgetManager when instantiating the widget.

PR #750, opened 2020-05-03
Fix bdf file writer

daniellasry:
Wasn't too bad after all.

Main part of the change was updating writeChannelDataValues() and writeAuxDataValues() to take a double[][] as arguments and write the data in the same format. I had to convert from doubles back to an array of 3 bytes.

### DIFF
--- a/OpenBCI_GUI/DataLogger.pde
+++ b/OpenBCI_GUI/DataLogger.pde
@@ -7,6 +7,10 @@ class DataLogger {
 
     }
 
+    ~DataLogger() {
+        closeLogFile();
+    }
+
     public void initialize() {
 
     }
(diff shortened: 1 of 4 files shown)


OpenBCI_GUI/DataLogger.pde
```diff
@@ -107,14 +107,14 @@ class DataLogger {
...
     */
     private void openNewLogFileBDF(String _fileName) {
-        if (fileoutput_bdf != null) {
+        if (fileWriterBDF != null) {
             println("OpenBCI_GUI: closing log file");
             closeLogFile();
         }
         //open the new file
-        fileoutput_bdf = new OutputFile_BDF(getSampleRateSafe(), nchan, _fileName);
+        fileWriterBDF = new DataWriterBDF(_fileName);
```
retiutut, line 115: I like the new name for this method.

PR #753, opened 2020-05-05
Basic popup window

daniellasry:
Usage example:
`PopupMessage msg = new PopupMessage("Info", "Streaming needs to be stopped before accessing hardware settings");`

image

### DIFF
--- a/OpenBCI_GUI/PopupMessage.pde
+++ b/OpenBCI_GUI/PopupMessage.pde
@@ -1,0 +1,109 @@
+import java.awt.Frame;
+import processing.awt.PSurfaceAWT;
+import processing.awt.PSurfaceAWT.SmoothCanvas;
+
+// Instantiate this class to show a popup message
+
+class PopupMessage extends PApplet {
+    private final int defaultWidth = 500;
+    private final int defaultHeight = 250;
+
+    private final int headerHeight = 55;
+    private final int padding = 20;
+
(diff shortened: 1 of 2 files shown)


retiutut:
I'm not a fan of pop-ups, but:
- style is consistent w/ existing styles
- not much code
- could be useful in the future when the `output()` method just isn't enough (messages at bottom of gui)

daniellasry:
> I'm not a fan of pop-ups, but:
> 
> * style is consistent w/ existing styles
> * not much code
> * could be useful in the future when the `output()` method just isn't enough (messages at bottom of gui)
> 
> Another alternative is to make this an overlay in the existing window (like "Starting Session" overlay added in 4.2.0), but that might produce more spaghetti code.

The first use case of this is to alert users that they need to stop stream before changing hardware settings,
- output() is not enough. Users will wonder why the button is not working and not notice the message.
- overlaying like "starting session" is not what we want, we do not want to stop drawing everything else. And even if we still drew everything else in the background, we would have to deal with buttons clicking though and all sorts of nasty stuff. And it's not any better than a pop up becasuse you still have to dismiss it the same way.
- I'm not sure what you have against pop ups but it's appropriate in this case.

The other viable alternative I see is to stop the stream when users go into hardware settings. But we discusses that we prefer not to mess with a user's session unexpectedly.

PR #755, opened 2020-05-06
Fix networking widget

daniellasry:
This is the minimum fix to glue things back together again. I did the minumum possible refactoring, and am not attempting to fix any of the many blatant issues that already existed in this widget (bad thread safety, lots of aux data is skipped, and the code is an embarassement).

`dataBufferToSend `is a replacement for `yLittleBuff_uV`
The `dataAccumulationQueue` is used to re-create the accumulation logic that the DataProcessing class used to do. Once enough data is in the queue, it is copied to `dataBufferToSend` and the newDataToSend flag is set to true.

It seems to work, I've tested with the OSC test python script which was the only one that ran successfully for me.


OpenBCI_GUI/W_Networking.pde
```diff
@@ -1283,42 +1322,42 @@ class Stream extends Thread {
 
     Boolean checkForData() {
```
retiutut, line 1323: This Boolean here.......... is my least favorite method in Networking, and I think it messes with the outgoing sample rate.

OpenBCI_GUI/W_Networking.pde
```diff
@@ -1786,8 +1831,7 @@ class Stream extends Thread {
                     msg.clearArguments();
                     msg.add(i+1);
                     //ADD Accelerometer data
-                    msg.add(hub.validAccelValues[i]);
-                    // println(i + " | " + hub.validAccelValues[i]);
+                    msg.add((int)lastSample[analogChannels[i]]);
```
retiutut, line 1834: Awesome, thanks for the fix here!

retiutut:
I still need to test this, but it looks good enough to merge for now, as this is not a comprehensive fix for W_Networking's shortcomings.

PR #756, opened 2020-05-06
Add windows signing to automatic build

daniellasry:
I also had to renew our certificates.

I'll squash this one.


daniellasry:
I jsut figure oud that secure variables are not exposed in PR builds. Which makes sense for security... Have to disable signing in PR builds

PR #759, opened 2020-05-07
Cleanup dead code

daniellasry:
I started removing DataPacket_Ads1299 and when is tarted pulling this is everything that came with it.

Removed CytonLegacy and MarkerMode

Fixed the pulse sensor widget


OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -1066,53 +1006,3 @@ void drawOverlay() {
...
-            loc = getWindowLocation(P2D);
-            appletOriginX = (int)loc.x;
-            appletOriginY = (int)loc.y;
-            windowOriginSet = true;
-            mouseInFrame = true;
-        }
-    }
-}
-
-PVector getWindowLocation(String renderer) {
```
retiutut, line 1104: Removing this is fine, since we use P2D to render the GUI.

daniellasry: removing this is fine because it was not called from anywhere

PR #760, opened 2020-05-07
Add Config Options for NovaXR in Control Panel

retiutut:
Screen Shot 2020-05-07 at 4 40 54 PM

Just the front-end solution in this PR.

Another aspect of this PR: proof-of-concept that we can put Buttons and other UI elements as `private` inside the "Box" classes in control panel, instead of letting them be global.

### DIFF
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -233,6 +233,7 @@ public void controlEvent(ControlEvent theEvent) {
         Map bob = ((ScrollableList)theEvent.getController()).getItem(val);
         // this will retrieve the enum object stored in the dropdown!
         novaXR_boardSetting = (NovaXRMode)bob.get("value");
+        println("ControlPanel: User selected NovaXR Mode: " + novaXR_boardSetting.getName());
     }
 
     //This dropdown is in the SessionData Box
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -142,7 +142,7 @@ public void controlEvent(ControlEvent theEvent) {
         protocolWifiCyton.setColorNotPressed(colorNotPressed);
         protocolSerialCyton.setColorNotPressed(colorNotPressed);
 
-        //Reset protocol and hide all cp5 elements
+        //Reset protocol
         selectedProtocol = BoardProtocol.NONE;
 
         //Perform this check in a way that ignores order of items in the menulist
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -227,6 +227,16 @@ public void controlEvent(ControlEvent theEvent) {
         }
     }
 
+    //Check for event in NovaXR Mode List in Control Panel
+    if (novaXREnabled) {
+        if (theEvent.isFrom("novaXR_Modes")) {
+            Map bob = ((ScrollableList)theEvent.getController()).getItem(int(theEvent.getValue()));
+            String s = (String)bob.get("text");
+            println("ControlPanel: Selected NovaXR Board Mode: " + s);
+            //todo[brainflow] set config_board message here, send message on init session
+        }
+    }
+
     if (theEvent.isFrom("maxFileDuration")) {
         int n = (int)theEvent.getValue();
         settings.setLogFileDurationChoice(n);
@@ -866,6 +876,10 @@ class ControlPanel {
                 }
             }
 
+            else if (novaXREnabled && eegDataSource == DATASOURCE_NOVAXR) {
+                novaXRBox.mousePressed();
+            }
+
             
             //The following buttons apply only to Cyton and Ganglion Modes for now
             if (autoSessionName.isMouseHere()) {
@@ -1163,6 +1177,10 @@ class ControlPanel {
                         new File(settings.guiDataPath + 
                                 "Sample_Data" + System.getProperty("file.separator") + 
                                 "OpenBCI-sampleData-2-meditation.txt"));
+        }
+
+        if (novaXREnabled) {
+            novaXRBox.mouseReleased();
         }
 
         //reset all buttons to false
@@ -2133,7 +2151,9 @@ class RecentPlaybackBox {
         textAlign(LEFT, TOP);
         text("PLAYBACK HISTORY", x + padding, y + padding);
         popStyle();
-        cp5_recentPlayback_dropdown.get(ScrollableList.class, "recentFiles").setVisible(true);
+        if (!cp5_recentPlayback_dropdown.get(ScrollableList.class, "recentFiles").isVisible()) {
+            cp5_recentPlayback_dropdown.get(ScrollableList.class, "recentFiles").setVisible(true);
+        }
         cp5_recentPlayback_dropdown.draw();
     }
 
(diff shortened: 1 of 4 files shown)


retiutut:
Need to update Mode dropdown to include additional modes.

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -227,6 +227,16 @@ public void controlEvent(ControlEvent theEvent) {
         }
     }
 
+    //Check for event in NovaXR Mode List in Control Panel
+    if (novaXREnabled) {
+        if (theEvent.isFrom("novaXR_Modes")) {
+            Map bob = ((ScrollableList)theEvent.getController()).getItem(int(theEvent.getValue()));
+            String s = (String)bob.get("text");
+            println("ControlPanel: Selected NovaXR Board Mode: " + s);
+            //todo[brainflow] set config_board message here, send message on init session
```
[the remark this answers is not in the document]
retiutut: Got it, save the value here, send message on init. Just got the values in today, so I can make the change.

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -2133,7 +2151,9 @@ class RecentPlaybackBox {
         textAlign(LEFT, TOP);
         text("PLAYBACK HISTORY", x + padding, y + padding);
         popStyle();
-        cp5_recentPlayback_dropdown.get(ScrollableList.class, "recentFiles").setVisible(true);
+        if (!cp5_recentPlayback_dropdown.get(ScrollableList.class, "recentFiles").isVisible()) {
+            cp5_recentPlayback_dropdown.get(ScrollableList.class, "recentFiles").setVisible(true);
+        }
```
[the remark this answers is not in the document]
retiutut: Goes hand in hand with another change you requested here. Good find.

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -2242,12 +2273,21 @@ class NovaXRBox {
...
+        createDropdown("novaXR_Modes", Arrays.asList(novaXRModes));
+        novaXRcp5.get(ScrollableList.class, "novaXR_Modes").setPosition(x + padding, novaXR250.but_y + 24 + padding);
     }
 
     public void update() {
         //Check for state change so we don't call setVisible() every update
         if (isShowing != previousIsShowing) {
-            cp5.get(Textfield.class, "novaXR_IP").setVisible(isShowing);
+            novaXRcp5.get(Textfield.class, "novaXR_IP").setVisible(isShowing);
+            novaXRcp5.get(ScrollableList.class, "novaXR_Modes").setVisible(isShowing);
```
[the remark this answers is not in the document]
daniellasry: Offline discussion: We can probably delete this because novaXRcp5 drawing is already controlled byt the draw() function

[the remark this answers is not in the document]
retiutut: It seems we still need to tell cp5 if it an element should be visible. Also making an element visible causes it to draw even if the draw function is not being called?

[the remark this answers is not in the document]
retiutut: Screen Shot 2020-05-08 at 8 28 29 PM

I told the cp5 elements to be visible even when I click another data source, and still shows, despite draw() not being called.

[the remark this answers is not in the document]
daniellasry: Did a quick google, turn out we need to set this flag: http://sojamo.de/libraries/archive/controlP5-0-3-14/reference/controlp5_method_setautodraw.htm

OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -261,6 +261,8 @@ int numSettingsPerChannel = 6; //each channel has 6 different settings
 char[][] channelSettingValues = new char [nchan][numSettingsPerChannel]; // [channel#][Button#-value] ... this will incfluence text of button
 char[][] impedanceCheckValues = new char [nchan][2];
 
+char novaXR_boardSetting = 'd'; //default mode
```
retiutut, line 264: @Andrey1994 Here is the char variable that gets set by the front-end

Andrey1994: I send this char in brainflow in prepare_session call, no problem with it. But there is a huge problem which we will need to fix, since now we track gain in GUI side(old GUI doesnt) we need to implement it for novaxr and for all modes too. Right now tracking works in assumption that gain is 24 for all channels but its not true anymore....

Andrey1994: PR which added this feature(in assumption that gain is 24 everywhere) https://github.com/OpenBCI/OpenBCI_GUI/pull/752

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -228,13 +228,28 @@ public void controlEvent(ControlEvent theEvent) {
...
+            case 2:
+                novaXR_boardSetting = 'g';
+                break;
+            case 3:
+                novaXR_boardSetting = 'h';
+                break;
+            case 4:
+                novaXR_boardSetting = 'j';
+                break;
+        }
```
daniellasry, line 251: Actually, we don't want this here. By putting the commands here, we are leaking board implentation details into UI code. Also, we're making this code flaky and hard to modify. What guarantee do we have that the dropdown value, drop down names, and commands will stay in sync? Here's what to do instead.

Reap up about Enums if you're not too familiar with them. These links look pretty good: [one](https://howtodoinjava.com/java/enum/java-enum-string-example/), [two](https://howtodoinjava.com/java/enum/enum-tutorial/)

Declare an Enum in BoardNovaXR. This will be our "source of truth" mapping between mode ID, mode name, and mode command:
```
public enum NovaXRMode
{
    DEFAULT("Default Mode", "d"), 
    INTERNAL_SIGNAL("Internal Signal", "f"), 
    EXTERNAL_SIGNAL("External Signal", "g"), 
    PRESET4("Preset 4", "h"),
    PRESET5("Preset 5", "j");

    private String name;
    private String command;
 
    Environment(String _name, String _command) {
        this.name = _name;
        this.command = _command;
    }
 
    public String getName() {
        return name;
    }

    public String getCommand() {
        return command;
    }
}
```

Then you can populate the dropdown menu like this:
```
ScrollableList modeList = novaXRcp5.addScrollableList(name)
            .setOpen(false)
            .setColorBackground(color(31,69,110)) // text field bg color
            ...
            ...
            .setVisible(true)
            ;

// foe each entry in the enum, add it to the dropdown.
for (NovaXRMode mode : NovaXRMode.values()) {
    // this will store the *actual* enum object inside the dropdown!
    modeList.addItem(mode.getString(), mode);
}
```

Then your code above will look like:
```
if (theEvent.isFrom("novaXR_Modes")) {
        int val = (int)(theEvent.getController()).getValue();
        Map bob = ((ScrollableList)theEvent.getController()).getItem(val);
        // this will retrieve the enum object stored in the dropdown!
        novaXR_boardSetting = (NovaXRMode)bob.get("value");
}
```

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -142,8 +142,8 @@ public void controlEvent(ControlEvent theEvent) {
         protocolWifiCyton.setColorNotPressed(colorNotPressed);
```
daniellasry, line 142: Note: Not something to fix this change, but there's something wrong with all the controllers using the same 'controlEvent' function and having to check "theEvent.isFrom". I believe there is a way to register separate callback functions for each controller.

OpenBCI_GUI/OpenBCI_GUI.pde
daniellasry: @Andrey1994 That's not a huge problem I could address it in my HardwareSettings refactor

OpenBCI_GUI/ControlPanel.pde
Andrey1994: Yep, its correct. More examples for enums can be found here https://github.com/brainflow-dev/brainflow/blob/master/java-package/brainflow/src/main/java/brainflow/BoardIds.java (I just copypaste this template every time)

OpenBCI_GUI/ControlPanel.pde
retiutut: Sounds great! Again, this is a part of the code that is not my favorite, but I like the idea of simply checking events, instead of convoluted checks/balances (ex. Button class).

OpenBCI_GUI/ControlPanel.pde
daniellasry: FYI I have implemented "enum NovaXRMode" in my open PR. We'll have to take that into account depending on who merges first.

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -2346,25 +2332,31 @@ class NovaXRBox {
...
+            // .addItems(_items) // used to be .addItems(maxFreqList)
             .setVisible(true)
             ;
+        // for each entry in the enum, add it to the dropdown.
+        for (NovaXRMode mode : NovaXRMode.values()) {
+            // this will store the *actual* enum object inside the dropdown!
+            modeList.addItem(mode.getName(), mode);
+        }
+        //Style the text in the ScrollableList
         novaXRcp5.getController(name)
```
daniellasry, line 2356: Small detail. Now that you're saving the ScrollableList as modeList  (which I think is great). You don't have to call "novaXRcp5.getController(name)". You can just use "modeList" in its place.

retiutut: Nice, let me try it out real quick. Always down for shorter lines of code.

retiutut: Changes made!

daniellasry: Looks great!

PR #761, opened 2020-05-07
Gui 5 widget cleanup

retiutut:
Clean up Focus, SSVEP,  and Presentation mode

Piggy-backs on #759 from @daniellasry


OpenBCI_GUI/ControlPanel.pde
```diff
@@ -1985,7 +1985,7 @@ class SampleRateCytonBox {
         fill(bgColor); //set color to green
         textFont(h3, 16);
         textAlign(LEFT, TOP);
-        text("  " + str(getSampleRateSafe()) + "Hz", x + padding + 142, y + padding); // print the channel count in green next to the box title
+        text("  " + str(currentBoard.getSampleRate()) + "Hz", x + padding + 142, y + padding); // print the channel count in green next to the box title
```
Andrey1994, line 1988: its a little unclear to me how it works... I think in ControlPanel there is no currentBoard object(its set to BoardNull) and only when we close controlPanel we create correct instance

daniellasry: It doesn't work you are correct. That text field should simply be removed.
image

daniellasry: removed in upstream change

daniellasry: Deleted in upstream change

PR #765, opened 2020-05-12
Fix error in synthetic board that was causing 4 channels not to work.

daniellasry:


OpenBCI_GUI/BoardSynthetic.pde
```diff
@@ -157,52 +157,51 @@ class BoardSynthetic extends Board implements AccelerometerCapableBoard {
     private void synthesizeEXGData(double[][] buffer, int sampleIndex) {
         float val_uV;
         for (int i = 0; i<getNumEXGChannels(); i++) {
-            int Ichan = exgChannels[i];
             if (isEXGChannelActive(i)) {
                 val_uV = randomGaussian()*sqrt(getSampleRate()/2.0f); // ensures that it has amplitude of one unit per sqrt(Hz) of signal bandwidth
-                if (Ichan==0) {
+                if (i==0) {
```
retiutut, line 162: Thanks for replacing this with just `i`!

retiutut:
It works! Thanks!

PR #768, opened 2020-05-13
Start Cyton+SD Fix after Data Refactor

retiutut:
- Does everything but send the command!


OpenBCI_GUI/BoardCyton.pde
```diff
@@ -320,6 +348,12 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         currentBoardMode = boardMode;
     }
 
+    public void openSDFile() {
+        println("Opening SD file. Writing " + cyton_sdSetting.getCommand() + " to OpenBCI.");
+        configBoard(cyton_sdSetting.getCommand()); // tell the SD file to close if one is open...
+        delay(100); //make sure 'j' gets sent to the board
```
Andrey1994, line 354: I think we dont need to insert delays here, maybe it was impartant in HUB since its another process which receives and processes commands in uknown time but now I dont see necessity in it

Andrey1994: And I dont get how it works, looks like its a new method but I dont see a place where its called

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -320,6 +348,12 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
...
+    }
+
     public void closeSDFile() {
         println("Closing any open SD file. Writing 'j' to OpenBCI.");
         configBoard("j"); // tell the SD file to close if one is open...
```
Andrey1994, line 359: Here I think we also can use enums

daniellasry: I thin kthe command to close sd file is alwasys 'j' so not necessary?

OpenBCI_GUI/BoardCyton.pde
daniellasry: I also think we don't need the delay here

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -320,6 +348,12 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         currentBoardMode = boardMode;
     }
 
+    public void openSDFile() {
+        println("Opening SD file. Writing " + cyton_sdSetting.getCommand() + " to OpenBCI.");
+        configBoard(cyton_sdSetting.getCommand()); // tell the SD file to close if one is open...
```
daniellasry, line 353: tells the sd file to open, not close

OpenBCI_GUI/BoardCyton.pde
retiutut: Unfortunately, using `j` to close the file is not an option in the dropdown menu, so adding this to the enum would add it as an option on the front-end, unless i tell it to exclude this option. I'd rather just leave it here.

OpenBCI_GUI/BoardCyton.pde
retiutut: also need to put a check that NO_WRITE is **not** selected here.

OpenBCI_GUI/BoardCyton.pde
retiutut: Removing the delay here.

OpenBCI_GUI/BoardCyton.pde
daniellasry: Per my comment, it's fine the way it is.

OpenBCI_GUI/BoardCyton.pde
daniellasry: You could, although you have the NO_WRITE command set to null, which will not send to the board.

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -327,10 +355,56 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
...
+        streaming = false;
+        closeSDFile();
+        try {
+            boardShim.stop_stream ();
+        }
+        catch (BrainFlowError e) {
+            println("ERROR: Exception when stoppping stream");
+            e.printStackTrace();
+        }
+    }
```
daniellasry, line 393: You don't need to re-implement the whole function, just:

```
    @Override
    public void startStreaming() {
        super.startStreaming();
        if(streaming) {
            openSDFile();
        }
    }

    @Override
    public void stopStreaming() {
        super.stopStreaming();
        closeSDFile();
    }
```

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -188,15 +188,13 @@ public void controlEvent(ControlEvent theEvent) {
...
-            output("OpenBCI microSD Setting = " + sdSettingString + " recording time");
-        } else {
-            output("OpenBCI microSD Setting = " + sdSettingString);
-        }
-        verbosePrint("SD setting = " + sdSetting);
+        int val = (int)(theEvent.getController()).getValue();
+        Map bob = ((ScrollableList)theEvent.getController()).getItem(val);
+        cyton_sdSetting = (CytonSDMode)bob.get("value");
+        String outputString = "OpenBCI microSD Setting = " + cyton_sdSetting.getName();
+        if (val != 0) outputString += " recording time";
```
daniellasry, line 195: instead of if val != 0 you should check if (cyton_sdSetting != CytonSDMode.NO_WRITE)
(What if the order of the enum changes, or the mapping gets out of sync for some reason?)

daniellasry: and always use brackets

```
if (cyton_sdSetting != CytonSDMode.NO_WRITE) {
    outputString += " recording time";
}
```

PR #763, opened 2020-05-11
Hardwaresettings Re-write and NovaXR modes

daniellasry:
- Wrote Enums for each ADS settings
- Enum instance holds button name
- Gain enum instance also hold gain scalar
.
- Wrote class that contains all hardware settings for a board "ADS1299Settings"
- It has a list of enums per setting, per channel.
- It has functionality to send settings command to board
- ADS1299SettingsController can ask board for hardware settings instance, keep it, modify it, and ask it to commit settings to board.
.
- Created insterface for boards that have ADS1299 settings (ADS1299SettingsBoard )
.
- HardwareSettingsController re-written and re-named ADS1299SettingsController
- ADS1299SettingsController hold an instance of board settings object
- When button is pressed, ask settings object to increment the setting's enum and send the command to board.
- Button names are populated based on which enum values is selected in settings object.
.
- Cyton default settings and NovaXR modes are defined by hard-coding in Settings object constructor.
- NovaXRDefault mode constructor will send mode command to board.

PR #769, opened 2020-05-13
fix ganglion impedance widget

Andrey1994:


Andrey1994:
It uses new brianflow.jar(important for #763 )

OpenBCI_GUI/BoardGanglion.pde
```diff
@@ -130,7 +144,28 @@ abstract class BoardGanglion extends BoardBrainFlow implements AccelerometerCapa
     }
 
     public void setCheckingImpedance(boolean checkImpedance) {
-        configBoard(checkImpedance ? "z" : "Z");
+        // todo brainflow, we need to change button to start/stop streaming
```
Andrey1994, line 147: looks like we dont need to

Andrey1994:
I think its data sent by device. @daniellasry do you see any errors here https://github.com/brainflow-dev/brainflow/blob/master/src/board_controller/openbci/ganglion_wifi.cpp#L278 ?
Looking at raw bytes format is:
* START_BYTE
* Channel num
* ASCI String with int num
* Zeroes(no 'Z')

daniellasry:
> I think its data sent by device. @daniellasry do you see any errors here https://github.com/brainflow-dev/brainflow/blob/master/src/board_controller/openbci/ganglion_wifi.cpp#L278 ?
> Looking at raw bytes format is:
> 
> * START_BYTE
> * Channel num
> * ASCI String with int num
> * Zeroes(no 'Z')

Yeah this is incorrect: `std::string asci_value ((const char *)(b + 2), i - 2);`

Should be: `std::string asci_value ((const char *)(b + 2), i - 1);`

OpenBCI_GUI/BoardGanglion.pde
```diff
@@ -130,7 +144,28 @@ abstract class BoardGanglion extends BoardBrainFlow implements AccelerometerCapa
...
+        if (checkImpedance) {
+            if (isCheckingImpedance) {
+                println("Already checking impedance.");
+                return;
+            }
+            if (streaming) {
+                stopStreaming();
```
daniellasry, line 154: Instead of "stopStreaming()" here we need to call the global "stopRunning()"
All the other calls to start/stop are fine. Just this one needs to change.

This will fix a bug in this change where if you click "Start Impedance" while the stream is running, after you stop impedance and click "Start Data Stream" nothing will happen.

(A better solution long term you be to get rid of the "isRunning" variable and replace it with currentBoard.isStreaming() ). And also get rid of the "startRunning()" and "stopRunning()" global functions and have the dataLogger be a listener on the board's "onStartStreaming" and "onStopStreaming" events. But i'm not expecting this kind of refactor in this PR)

OpenBCI_GUI/W_GanglionImpedance.pde
```diff
@@ -71,11 +78,13 @@ class W_GanglionImpedance extends Widget {
...
                 popStyle();
             }
-            // Display a spinning gif if impedance check is on
+
             image(loadingGIF_blue, x + padding + startStopCheck.but_dx + 15, y + padding - 8, 40, 40);
+            popStyle();
+        } catch (Exception e) {
+            e.printStackTrace();
+            return;
         }
```
daniellasry, line 87: I don't like big try/catches that catch every exception. Do you even need this? What bug is it hiding?

Andrey1994: I dont like it either but I dont understand processing.... Looks like this method is called in the beginning when board even is not selected and smth strange happens

daniellasry: I tried removing it locally and i had no issues at all.

Andrey1994: I have an exception in this line during processing start
```
 int resistanceChannels[] = BoardShim.get_resistance_channels(ganglion.getBoardIdInt());
```

daniellasry: You need to make a function in BoardGanglion "getImpedanceChannels()" that works similarly to "getAccelerometerChannels()".

Processing is complaining at compile time that "get_resistance_channels" throws a BrainflowError esception and you are forced to handle it.

Andrey1994: @daniellasry done

retiutut:
The values look correct now for Ganglion+WiFi. Tested on Mac.

Screen Shot 2020-05-19 at 12 31 23 AM

PR #773, opened 2020-05-19
Fix Widget Selector Drawing Error

retiutut:
- **Comment out `delay(1000)` in WidgetManager.pde**

- Appropriately size the widget selector at the right times to remove extra grey space being drawn.

- Minor refactor of Widget.pde to include appropriate access modifiers

Before, only looked right when widget dimensions were recalculated. Now, looks right the first time, and every time.

Test WidgetSelector:
- Clear all settings (only happens with default settings)
- Open GUI synthetic mode or Cyton mode
- Don't resize the GUI or move any widgets
- Hover over the Widget Selector dropdown in the TimeSeries (full height container)
- Check that there is no grey space

Test Widget Refactor:
- Aggressively test the WidgetSelector dropdown for all widgets in multiple containers

_____

**Before**
Screen Shot 2020-05-18 at 9 25 37 PM

**After**
Screen Shot 2020-05-18 at 10 21 00 PM


retiutut:
> I see many variables and methods are protected now, but I dont see the usage of them in subclasses. Do we really need to make them protected?

Maybe a few more `protected` vars can be `private` instead, I'll take another look. 🧩

OpenBCI_GUI/Widget.pde
```diff
@@ -10,28 +10,30 @@
...
 
     //some variables for the dropdowns
-    int navH = 22;
-    int widgetSelectorWidth = 160;
-    int dropdownWidth = 64;
+    protected final int navH = 22;
+    private int widgetSelectorWidth = 160;
+    private int widgetSelectorHeight = 0;
+    protected final int dropdownWidth = 64;
+    private boolean initialResize = false; //used to properly resize the widgetSelector when loading default settings
```
daniellasry, line 36: I appreciate that you're adding private/protected/public modified whenever you encounter them

OpenBCI_GUI/WidgetManager.pde
```diff
@@ -199,7 +199,7 @@ class WidgetManager{
             setNewContainerLayout(currentContainerLayout); //sets and fills layout with widgets in order of widget index, to reorganize widget index, reorder the creation in setupWidgets()
         }
 
-        delay(1000);
+        //delay(1000);
```
daniellasry, line 202: Just delete it! I don't see it coming back.

retiutut: I asked Conor about this today, and he said this may have been put in place when widget instantiation was non-optimized. I suspect that the data refactor has reduced the strain on Session Start, and that this is no longer needed.

daniellasry: I'm not even sure what the purpose of waiting was? There's nothing going on on a separate thread. My best guess is that it was added to fix a bug that was not well understood, instead of fixing the actual bug.

We won't be doing this kind of thing moving forward :)

OpenBCI_GUI/Widget.pde
retiutut: dropdownWidth can be private

PR #774, opened 2020-05-20
Fix mac release crashing on launch

daniellasry:
Revert to processing 3.5.3 (on all platforms to be consistent)


OpenBCI_GUI/Info.plist.tmpl
```diff
@@ -23,7 +23,7 @@
         <key>CFBundleShortVersionString</key>
         <string>4</string>
         <key>CFBundleVersion</key>
-        <string>5.0.0-alpha.5</string>
+        <string>5.0.0-alpha.6</string>
```
retiutut, line 26: Thank you for bumping the version!

PR #776, opened 2020-05-21
Take a pass at //todo[brainflow]

daniellasry:


OpenBCI_GUI/W_NovaAux.pde
```diff
@@ -83,19 +79,15 @@ class W_NovaAux extends Widget {
     public void setVisible(boolean _visible) {
         visible = _visible;
     }
-    public void setUpdating(boolean _updating) {
-        updating = _updating;
-    }
 
     void update() {
-        if(visible && updating) {
+        if(visible) {
```
retiutut, line 84: Since you followed the same pattern as other widgets, this should be correct. Other widgets have been tested to update/draw correctly and they are running without exceptions.

OpenBCI_GUI/W_Playback.pde
```diff
@@ -58,16 +57,10 @@ class W_playback extends Widget {
     public boolean isVisible() {
         return visible;
     }
-    public boolean isUpdating() {
```
retiutut, line 61: Playback widget still works with this change.

PR #778, opened 2020-05-23
GUI v5 Networking Repairs

### LINKED ISSUE #644: Remove #Chan Textfield from LSL
## Problem

Num Chan parameter in Networking Widget is really unclear and weird. Other datatypes seem to do just fine without this option.

## Expected 

Send data over LSL without specifying some arbitrary and inconsistent value. Plus, the name "#Chan" does't apply to all datatypes. Let's remove it and make it more like the interface for the other three datatypes. 🛠 

All OS

## GUI Version

4.1.7-beta.0

## Running standalone app

Both

## Type of OpenBCI Board

All boards

No

## Console Log

@sentdex Reports that setting #Chan to 125 for Cyton+Daisy FFT datatype works instead of 16. So, lets just remove the feature altogether and let the GUI figure it out.

retiutut:
- [x] Remove #Chan Textfield from Networking - Closes #644 
- [x] save networking filters as booleans to JSON
- [x] simplify networking filter buttons
- [x] replace filter radio buttons with cp5 Toggles 
- [x] display "on" or "off" next to filter buttons aligning with current state
- [x] refine/troubleshoot filter buttons (disable when data type dropdown is open)

https://github.com/OpenBCI/OpenBCI_GUI/pull/778/commits/d9f74b3fdfd1570a6e2debacca258bae2860f730 - The GUI `Button` class was blocking access to Cp5 `Button` class, this commit renames all GUI Buttons to `Button_obci`, and it will be easier to convert them to cp5 buttons at another time.

### DIFF
--- a/OpenBCI_GUI/W_Networking.pde
+++ b/OpenBCI_GUI/W_Networking.pde
@@ -100,13 +100,13 @@ class W_Networking extends Widget {
         "127.0.0.1","12347",
         "127.0.0.1","12348"};
     String[] lslTextFieldNames = {
-        "LSL_name1","LSL_type1","LSL_numchan1",
-        "LSL_name2","LSL_type2","LSL_numchan2",
-        "LSL_name3","LSL_type3","LSL_numchan3"};
+        "LSL_name1","LSL_type1",
+        "LSL_name2","LSL_type2",
+        "LSL_name3","LSL_type3"};
     String[] lslTextDefaultVals = {
-        "obci_eeg1","EEG",Integer.toString(nchan),
-        "obci_eeg2","EEG",Integer.toString(nchan),
-        "obci_eeg3","EEG",Integer.toString(nchan)};
+        "obci_eeg1","EEG",
+        "obci_eeg2","EEG",
+        "obci_eeg3","EEG"};
     String networkingGuideURL = "https://openbci.github.io/Documentation/docs/06Software/01-OpenBCISoftware/GUIWidgets#networking";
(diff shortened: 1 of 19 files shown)


retiutut:
Proof of concept using CP5 that have unique identifiers,  but also look the same and have the same label.

Screen Shot 2020-05-25 at 1 29 45 PM

retiutut:
Accidentally did this, and made me rethink how to do this: Just use a Cp5 button with label so I don't have to worry about drawing "on" or "off" text in a certain position.

Screen Shot 2020-05-25 at 1 31 48 PM

retiutut:
Screen Shot 2020-05-25 at 5 12 50 PM

OpenBCI_GUI/W_Networking.pde
```diff
@@ -49,18 +49,22 @@ class W_Networking extends Widget {
     int row5;
     int itemWidth = 96;
     private final float datatypeDropdownScaling = .35;
+    private final int filtButW = 40;
+    private final int filtButH = 20;
+    private final int filtOffsetX = (itemWidth / 2) - (filtButW / 2);
```
retiutut, line 54: This should not be final, and should be recalculated when the Widget gets resized.

retiutut: Done!

retiutut:
Screen Shot 2020-05-26 at 1 39 14 PM
Screen Shot 2020-05-26 at 1 39 32 PM
Resizing Test : Pass ✅

PR #779, opened 2020-05-26
Repair Spectrogram after Data Refactor

retiutut:
- Remove "Save Spectrogram" button - not enough room, and GUI screenshot is better idea anyways
- Fix timestamps

### DIFF
--- a/OpenBCI_GUI/W_Spectrogram.pde
+++ b/OpenBCI_GUI/W_Spectrogram.pde
@@ -18,7 +18,6 @@ class W_Spectrogram extends Widget {
     int xPos = 0;
     int hueLimit = 160;
 
-    Button widgetTemplateButton;
     PImage dataImg;
     int dataImageW = 1800;
     int dataImageH = 200;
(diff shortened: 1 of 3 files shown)


retiutut:
Let me see if I can update the code. The timestamp here is only used to display on the screen.

OpenBCI_GUI/W_Spectrogram.pde
```diff
@@ -464,16 +425,14 @@ class W_Spectrogram extends Widget {
...
+        List<double[]> currentData = currentBoard.getData(1);
+        int timeStampChan = currentBoard.getTimestampChannel();
+        long timestampMS = (long)(currentData.get(0)[timeStampChan] * 1000.0);
+        if(timestampMS == 0) {
+            return "TimeNotFound";
         }
-        
-        return "TimeNotFound";
+        String output = new SimpleDateFormat("HH:mm:ss").format(timestampMS);
+        return output;
```
daniellasry, line 435: Agreed with andrey. This function should not return a string, it should return a long. You can use your `DateTimeFormatter formatter` above to format it later.

OpenBCI_GUI/W_Spectrogram.pde
```diff
@@ -429,27 +396,21 @@ class W_Spectrogram extends Widget {
...
+                time = LocalTime.parse(getCurrentTimeStamp());
             }
+        } catch (NullPointerException e) {
+            println("Spectrogram: Timestamp error...");
+            e.printStackTrace();
+            time = LocalTime.now();
+        } catch (NumberFormatException e) {
+            println("Spectrogram: Timestamp error...");
+            e.printStackTrace();
+            time = LocalTime.now();
```
daniellasry, line 413: Instead of catching these exceptions, fix the actual bug they are hiding.

retiutut: I think the data refactor + brainflow has mitigated these exceptions.

retiutut:
Switching usage of `LocalTime` to `LocalDateTime` since the latter includes the following method, allowing use to input a `long` from currentBoard:

https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#ofInstant-java.time.Instant-java.time.ZoneId-

PR #780, opened 2020-05-26
Lsl timestamps fix

### LINKED ISSUE #775: LSL manual timestamping interferes with LSL clock_offset correction
## Problem

Hi. I'm one of the LSL maintainers. I'm trying to help a user with a problem. They are unable to synchronize OpenBCI streams with those from different sources when loading xdf files saved in LabRecorder.

The xdf importer adjusts the timestamps in each stream to a common clock: lsl's local_clock on the computer that was running LabRecorder. This adjustment relies on a series of clock_offsets stored alongside the stream. The clock_offsets are calculated online and stored every ~5 seconds using the LabRecorder computer's and the **remote computer's lsl::local_clock**.
When the remote stream's timestamps also uses lsl::local_clock, then synchronizing the streams on data import 'just works'.

When the remote streams timestamps are set manually, as you might want to do if you had information from the hardware about sample_acq_time which you know to be superior to the default sample_push_time, then it's important to convert sample_acq_time into the same base clock as lsl_local_clock. 
If a different base clock is used then post-hoc synchronization of different LSL streams is impossible.

FYI, [lsl's clock is](https://github.com/sccn/liblsl/blob/a6966233cfd4d2fca52a8f75800d73a624cd4a9f/src/common.cpp#L22-L26) `std::chrono::high_resolution_clock::now().time_since_epoch()`.

Another important details is that the timestamps should be in seconds. When using `push_chunk`, timestamps for samples after the first sample in a chunk are auto-incremented based on the stream's nominal_rate as +1/rate **seconds**. If millisecond timestamps are used then the first timestamp in a chunk will be in milliseconds but subsequent timestamps in that chunk will be incremented in seconds. This would have the effect of all timestamps within a chunk being very close in time, and then relatively large gaps between chunks. This will interfere with xdf importer's ability to dejitter timestamps.

## Expected 

As it appears the manual timestamps are simply the time at sample push and do not make use of any hardware timing info, there's no benefit to supplying the timestamps manually. It would be simpler to not provide timestamps at all and let LSL timestamp them for you. This will enable post-hoc synchronization of different streams and fix the chunk timestamping error above.

The downside to this is that it might break compatibility with other software you might have that expects the LSL timestamps to be `System.currentTimeMillis()`. For example, if any of your example Matlab or Python scripts for reading LSL streams expects timestamps in msec then these will have to be adjusted.

Edit: I took a look at the examples and none that I saw made explicit use of timestamps so it shouldn't be an issue.

retiutut:
Closes #775

### DIFF
--- a/OpenBCI_GUI/BoardNovaXR.pde
+++ b/OpenBCI_GUI/BoardNovaXR.pde
@@ -4,49 +4,13 @@ import org.apache.commons.lang3.ArrayUtils;
 
 final boolean novaXREnabled = true;
 
-interface NovaXRSettingsEnum {
-    public String getName();
-    public String getCommand();
-}
-
-public enum NovaXRSR implements NovaXRSettingsEnum
-{
-    SR_250("250Hz", "~6", 250),
-    SR_500("500Hz", "~5", 500),
-    SR_1000("1000Hz", "~4", 1000);
-
-    private String name;
-    private String command;
-    private int value;
- 
-    NovaXRSR(String _name, String _command, int _value) {
-        this.name = _name;
-        this.command = _command;
-        this.value = _value;
-    }
- 
-    @Override
-    public String getName() {
-        return name;
-    }
-
-    @Override
-    public String getCommand() {
-        return command;
-    }
-
-    public int getValue() {
-        return value;
-    }
-}
-
-public enum NovaXRMode implements NovaXRSettingsEnum
+public enum NovaXRMode
 {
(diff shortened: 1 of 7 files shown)


daniellasry:
This change still in progress?

retiutut:
#775 has detailed information and states the above fix should be all that is necessary.

daniellasry:
I see some changes to spectrogram mixed-in here. I think you need to revert them

retiutut:
Ok done.

PR #781, opened 2020-05-27
add file converter

Andrey1994:


Andrey1994:
I am not sure that it works correctly but data in file looks ok and at least it can be loaded to new GUI...

I have no idea if rendering for EEG is fine but for accel in BrainFlow there is a repeat of last accel value if current value is 0 while here there is no such repeat(right now) so it looks like a saw

retiutut:
Thanks for getting this started! I'll take a look soon and test.

retiutut:
Screen Shot 2020-05-27 at 9 26 47 PM

retiutut:
Screen Shot 2020-05-28 at 9 10 46 PM

Something is wrong with the sample rate when playing back converted files.

retiutut:
Files to test with:
[OldGUI - Cyton8ch Accel.txt](https://github.com/OpenBCI/OpenBCI_GUI/files/4699333/OldGUI.-.Cyton8ch.Accel.txt)

[OldGUI - Cyton8ch Digital.txt](https://github.com/OpenBCI/OpenBCI_GUI/files/4699335/OldGUI.-.Cyton8ch.Digital.txt)


[OldGUI - Cyton 8Ch Analog.txt](https://github.com/OpenBCI/OpenBCI_GUI/files/4699332/OldGUI.-.Cyton.8Ch.Analog.txt)

Andrey1994:
I've fixed accel in playback files and GUI

Andrey1994:
For digital and analog data I think we just need to accept that it will not work, at least because now we need another data....

And I doubt that there will be a lot of requests for this converter even for accel mode

retiutut:
@Andrey1994 Sounds like a plan! Let's wait to hear from users since this is easy for you to change later.

Personal would-be-nice:
- Batch conversion
- command line argument for cyton accel vs digital vs analog, since the user should already know this, or find out

retiutut:
Fix coming now...

Andrey1994:
added space to sampling rate line

PR #784, opened 2020-05-30
add bufferedstack and make data stream look smooth

Andrey1994:

@retiutut Can you add UI controller to turn it off and on? We will need to pass value to Cyton constructor and call 

```
super(30, 250)
```
to make data stream "smooth" and call
```
super()
``` 
by default to plot real data

Also it would be great if you both test it, I've tested it on windows and set small and big latency for FTDI buffer to check.

Solution here is not perfect but its simple and does the job I think.

### DIFF
--- a/OpenBCI_GUI/FixedStack.pde
+++ b/OpenBCI_GUI/FixedStack.pde
@@ -1,8 +1,9 @@
 import java.util.Stack;
-
+import java.util.concurrent.locks.ReentrantLock;
 
 public class FixedStack<T> extends Stack<T> {
-    private int maxSize;
+
+    protected int maxSize;
 
     public FixedStack(int size) {
         super();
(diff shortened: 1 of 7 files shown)


daniellasry:
Nice solution! One thing that would make this event simpler is to have instances of both FixedStack and BufferedStack. Data can be added to both. Then we can have a toggle button (while streaming) to toggle between the two buffers. 

We would probably switch between the two stacks in the Board::getData() function.

Another thing: If you create the instance of BufferedStack in initialize(), you can call Board::GetSampleRate() and pass it to the BufferedStack constructor. Then you don't need hardcoded sample rates.

The FPS value we can probably hardcoded to 60? That way we don't force individual boards to provide hardcoded FPS and sampling rate, and they can go back to having 1 constructor.

Alternatively, you could BufferedStack run on the main thread. Make an update() function on it and release samples on every update.

Thanks!

OpenBCI_GUI/FixedStack.pde
```diff
@@ -32,4 +33,102 @@ public class FixedStack<T> extends Stack<T> {
...
+        while (internalStack.size() >= maxSize) {
+            internalStack.remove(0);
+        }
+        return internalStack.push(object);
+    }
+
+    @Override
+    public void run() {
+        int sleepTime = 1000 / fps;
+        int packagesPerFrame = (samplingRate * fps) / 1000;
```
Andrey1994, line 86: just realized that this equation is wrong should be samplingRate/fps @daniellasry is it correct?

Andrey1994:
<< Nice solution! One thing that would make this event simpler is to have instances of both FixedStack and BufferedStack. Data can be added to both. Then we can have a toggle button (while streaming) to toggle between the two stacks.

I thought about it and decided that more likely this switch will be in control panel and valid for entire session(its easier). Also, multiple buffers to store exactly the same data reminds me old code in dataProcessing.pde...

<<Another thing: If you create the instance of BufferedStack in initialize(), you can call Board::GetSampleRate() and pass it to the BufferedStack constructor. Then you don't need hardcoded sample rates.

You are right, but we will need to pass info about which buffers to use and info like sampling rate/fps to initialize method and we will spoil common interface with information which makes sense only for few boards(Cyton Serial)

<<The FPS value we can probably hardcoded to 60? That way we don't force individual boards to provide hardcoded FPS and sampling rate, and they can go back to having 1 constructor.

I would hardcode 30(in old gui its 20 as I remember), BufferedStack has a constructor with default fps

<<Alternatively, you could BufferedStack run on the main thread. Make an update() function on it and release samples on every update.

Yep I considered option to inform bufferedstack that frame was updated and avoid multithreading, but I've choosed this option

Andrey1994:
I think this hack which is a little bit scam in my opinion should be available\used not for all boards... I suppose it can be a checkbox only for cyton serial, thats why its in constructor instead initialize methods and thats why there is single buffer

daniellasry:
> I think this hack which is a little bit scam in my opinion should be available\used not for all boards... I suppose it can be a checkbox only for cyton serial, thats why its in constructor instead initialize methods and thats why there is a single buffer

I agree with that. Here's another idea. You could isolate the change to Cyton Serial only without much work:

We currently have `BoardCytonSerial` and `BoardCytonSerialDaisy` classes. You could make a `BoardCytonSerialBase` parent class. (just like we did for Cyton Wifi).

Move the BufferedStack instance to `BoardCytonSerialBase`. Override `getNewDataInternal()`. Call super.getNewDataInternal(), append this data to the BufferedStack, then ask the BufferedStack for the data to be delivered this frame, return that data.

Andrey1994:
<<Move the BufferedStack instance to BoardCytonSerialBase. Override getNewDataInternal(). Call super.getNewDataInternal(), append this data to the BufferedStack, then ask the BufferedStack for the data to be delivered this frame, return that data.

Unfortunately, not. There is a Board class which creates an instance of FixedStack and in fact I will need to override all methods....

daniellasry:
It would work fine. The board class would have a FixedStack and the BoardCytonSerial class you have a BufferedStack insternally. You would use the BufferedStack in getNewDataInternal(), to deliver data smoothly to the Board class. You would not need to change the Board class at all.

Andrey1994:
Yes, it will but I will need to override not one method but at least 3: getData, update and initialize

daniellasry:
Why do you say that? You only need to override getNewDataInternal(). Roughly like this:

(Class BoardCytonSerialBase)

```
@Override
    protected double[][] getNewDataInternal() {
        double[][] frameData = super.getNewDataInternal();
        myBufferedStack.append(frameData);
        double[][] smoothFrameData = myBufferedStack.getSmoothData();
        return smoothFrameData ;
    }
```

All you need is the code above, and an implementation of BufferedStack

daniellasry:
I guess you could say it's more of a queue than a stack

Andrey1994:
getData is called from widgets, getData implementation

```
@Override
    public List<double[]> getData(int maxSamples) {
        int endIndex = accumulatedData.size();
        int startIndex = max(0, endIndex - maxSamples);

        return accumulatedData.subList(startIndex, endIndex);
    }
```
So I will need to change it to return data from another buffer.

Also I need to push data to another buffer, so I need to override update also. And initialize should be differet

daniellasry:
No you don't need to do that... 

accumulatedData is populated by calling getNewDataInternal()
You are overriding getNewDataInternal() to deliver data smoothly
accumulatedData  will be populated smoothly
getData() will work fine

daniellasry:
You can push data into the BufferedStack in BoardCytonSerialBase::getNewDataInternal() and pull data from the BufferedStack immediately after. It does not have to be threaded.

Andrey1994:
Nice, will try

PR #785, opened 2020-05-31
Fix Playback Mode Sample Rate

retiutut:


OpenBCI_GUI/DataSourcePlayback.pde
```diff
@@ -55,7 +55,7 @@ class DataSourcePlayback implements DataSource, AccelerometerCapableBoard, Analo
             // some boards have configurable sample rate, so read it from header
             if (line.startsWith("%Sample Rate")) {
                 int startIndex = line.indexOf('=') + 2;
-                int endIndex = line.indexOf("Hz") - 1;
```
retiutut, line 58: Here is the change.

retiutut:
I also see that Travis build is failing? Doesn't seem like it is caused by this PR.

daniellasry:
> I also see that Travis build is failing? Doesn't seem like it is caused by this PR.

Could you go into travis.yml and replace all instances of `https` with `http`? That should fix it!

retiutut:
Thanks, will try next.

OpenBCI_GUI/DataSourcePlayback.pde
daniellasry: What was the bug? Was it failing to parse properly because if the space before `Hz`?

retiutut: Yep. Added a println and the sample rate was 25 instead of 250.

daniellasry: That's weird. If you decrease the end index it shoudl make it worse, not better. Are you sure this fixes the problem?

retiutut: Yes, this increases the end index by removing ` - 1`.

retiutut: Hold on there's more going on here...

PR #787, opened 2020-05-31
Cyton Radio Config Refactor + Fixes

retiutut:
- [x] Make RadioConfig a class and remove unnecessary methods
- [x] Repair / safeguard all radio config buttons (always close port when done with a method)
- [x] Filter Cyton Serial port list in control panel to show only OpenBCI dongles
- [x] Also, only show .cu serial port on Mac
- [x] Test on Linux and Windows


retiutut:
To Test:
- Try all RadioConfig buttons in any practical order
- If something breaks, you can always press autoscan
- Sometimes, radio config box will say "please press autoscan one more time", and this is still spot on and a special case that is handled gracefully.

OpenBCI_GUI/RadioConfig.pde
```diff
@@ -0,0 +1,342 @@
...
+    //= Used to align Cyton and Dongle on the same radio channel, in case there is a mismatch.
+    public void scan_channels(RadioConfigBox rcConfig){
+        println("Radios_Config: scan_channels");
+        autoscanPressed = true;
+        if(serial_direct_board == null){
+            if(!connect_to_portName(rcConfig)){
+                return;
+            }
+        }
+        for(int i = 1; i < 26; i++){
```
Andrey1994, line 37: I know nothing about radio channels but what is 1 and 26?

retiutut: The Cyton+Dongle has up to 26 channels (like a walkie-talkie). The autoscan button tries to find what channel the Cyton is on so that it can align the channels for both the Cyton and the Dongle.

Old + Simple solution. No need to change this.

daniellasry: Would be nice if these numbers were named constants intead so it's more clear. Or could you at least add a comment?

retiutut: Will do!

daniellasry:
Where's the code that makes it so only "cu" shows and not tty?

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -610,25 +607,25 @@ class ControlPanel {
...
-        serialPorts = new String[Serial.list().length];
-        serialPorts = Serial.list();
-        serialList.items.clear();
-        for (int i = 0; i < serialPorts.length; i++) {
-            String tempPort = serialPorts[(serialPorts.length-1) - i]; //list backwards... because usually our port is at the bottom
-            serialList.addItem(makeItem(tempPort));
+        comPortBox.serialList.items.clear();
+        String name = "FT231X USB UART";
+        SerialPort[] comPorts = SerialPort.getCommPorts();
+        for (int i = 0; i < comPorts.length; i++) {
```
retiutut, line 619: @daniellasry Here is the alternate Serial library that allows me to filter the list based on the name of the product. In this case, all Cyton dongles are named "FT231X USB UART", and it shows up this way for all OS. 😄

daniellasry: Understood, thanks!

OpenBCI_GUI/RadioConfig.pde
retiutut: done

PR #789, opened 2020-06-01
make data smooth for cyton

Andrey1994:

Refactor #784 

Still need UI controller


retiutut:
Should the UI toggle be accessible on app start? also during session?

Andrey1994:
Only in init session and only for cyton serial

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -83,6 +101,61 @@ class BoardCytonSerialDaisy extends BoardCyton {
...
+            for (int j = 0; j < getTotalChannelCount(); j++) {
+                newEntry[j] = data[j][i];
+            }
+            buffer.addNewEntry(newEntry);
+        }
+        int numData = buffer.getDataCount();
+        if (numData == 0) {
+            return emptyData;
+        }
+        // transpose back
```
daniellasry, line 146: One thing we can do in the future (not this change) Is to make all of our API transposed.

getData(), getFrameData() and getNewDataInternal() could all return `List<double>`. It would make our interface more consistent anyway

OpenBCI_GUI/Buffer.pde
```diff
@@ -0,0 +1,27 @@
...
+public class Buffer<T> extends LinkedList<T> {
+
+    private int samplingRate;
+    private int fps;
+    private int dataPerUpdate;
+
+    Buffer(int fps, int samplingRate) {
+        this.fps = fps;
+        this.samplingRate = samplingRate;
+        dataPerUpdate = Math.max(samplingRate / fps, 1);
```
daniellasry, line 13: This is not going to work if the fps is not exactly '30'. We need to calculate how many samples to release based on how much time has passed since the last call.

daniellasry:
> Only in init session and only for cyton serial

I had a 1:1 meeting with conor, where he told me he would much rather have it be switchable during the session. 

I think this would be very easy to do. Just switch the "smoothData" boolean and clear the buffer. (yes, we will skip some data during the transition, but we don't care).

OpenBCI_GUI/Buffer.pde
Andrey1994: I dont think so, you can write any number here and it should not match real sampling rate. Like in old GUI 20 was hardcoded...

Andrey1994: More preciesly you can write any number here which is close or less than real FPS, if FPS is 60 and here we specify 30 everything will work fine due to
```
public int getDataCount() {
        return Math.min(this.size(), dataPerUpdate);
    }
```
If real FPS is 10 and here we use 30 delay will occur, but assuming that GUI dont render at 10 FPS we are save I think.

Also probably for 10 FPS you dont need to smooth data ...

I can pass current FPS instead time between updates but I didnt find a way to get it in GUI....

daniellasry: Yes, and old gui code was incorrect. That's why we still noticed FTDI choppyness despite the smoothing code. That code was wrong.

If the average FPS is above 30, the data will still be somewhat choppy because we will empty the buffer too fast.

If the average FPS is below 30, the buffer will grow indefinitely and the delay will get large and larger.

You don't need to get real fps from gui. you just need to store the time since the last call to `getDataCount` (you can use`millis()`). and calculate the number of samples based on the time elapsed.

Therre is another problem though. What if the board says it has a sample rate of 250 but real sample rate is lower? (due to packet loss or any other reasons). Then the buffer will grow and the delay will grow. That would be bad (and the old gui code was suceptible to this issue).

It's a little tricky. The best solution I have come up with is to limit the size of the buffer, and if the max size is exceeded, "release" the excess samples in addition to the number of samples that should be release on the current frame.

Andrey1994: <<If the average FPS is above 30, the data will still be somewhat choppy because we will empty the buffer too fast.

No, because we dont have to make them to match FPS exactly, it will still be smoother if its less than real sampling rate

<< If the average FPS is below 30, the buffer will grow indefinitely and the delay will get large and larger.
Agreed, and its a problem, so we can provide lower possible but decent FPS to constructor

<< You don't need to get real fps from gui. you just need to store the time since the last call to getDataCount (you can usemillis()). and calculate the number of samples based on the time elapsed.

There is no uniform way to get real FPS?

daniellasry: > No, because we dont have to make them to match FPS exactly, it will still be smoother if its less than real sampling rate

Let's say that GUI's average FPS is 60, and that FTDI is sending data 1 time per second. 
If the first half of the second, the data will come in smoothly (although too fast). In the second half of the second, no data will draw. So it will not look smooth.

> Agreed, and its a problem, so we can provide lower possible but decent FPS to constructor

I don't think that's good enough, because it is diminishing the problem instead of fixing it...
Do you want me to implement it? I think I have a pretty good idea of how, and I could just add a few commits to this PR. I also realize that I'm giving you more work, and I don't want you to get frustrated.

> There is no uniform way to get real FPS?
Actually there is. The grobal variable "frameRate" is provided by processing. However, it's an [approximate average of the last few frames](https://processing.org/reference/frameRate.html). Using the time since last frame will be more precise and keep that data smooth as the frame rate fluctuates.

Andrey1994: <<Let's say that GUI's average FPS is 60, and that FTDI is sending data 1 time per second.
In the first half of the second, the data will come in smoothly (although too fast). In the second half of the second, no data will draw. So it will not look smooth.

Not, here we specify 30 real fps is 60 and sampling rate is 250. In this code it will plot 8 packages per each 33ms(each second frame in terms of GUI's fps) and it will look pretty smooth despite the fact that real fps is 60.

I dont think that here we need 100% precision

Andrey1994: but it doesnt matter its simple to add fixed buffer size and provide current framerate\time

daniellasry: > Not, here we specify 30 real fps is 60 and sampling rate is 250. In this code it will plot 8 packages per each 33ms(each second frame in terms of GUI's fps) and it will look pretty smooth despite the fact that real fps is 60.

How so? Maybe I misunderstood your code? As far as I can tell, getNewDataInternal() is called every frame. So if the fps is 60, it will be called 60 times per second. So it will plot ~8 samples every ~16ms, when ir really shoudl be plotting 4 (250/60).

Andrey1994: <<when it really should be plotting 4 (250/60). 
This statement is wrong we dont have to plot smth at each frame at it still will look smooth

daniellasry: You're right, we don't have to plot something new at each frame, but your code does

daniellasry: Unless I'm misunderstanding. If so, whhich part of your code makes it plot every other frame if the framerate is 60?

Andrey1994: Technically its like its own FPS for timeseries widget which can be less than main one. In this widget it will look exactly like GUI has FPS 30 but plot will be smooth(since its smooth when FPS is really 30)

Andrey1994: And this is not true also " So it will plot ~8 samples every ~16ms," since this buffer is not preallocated it can not plot more data than there is. It will plot 8 samples each 32ms

Andrey1994: > You're right, we don't have to plot something new at each frame, but your code does

buffer is not preallocated, size of this buffer will be 0  and getDataCount will return 0 so we will skip one "main frame"

Andrey1994:
```
return Math.min(this.size(), dataPerUpdate);
```
maybe this should be change to smth like
```
if size >= dataPerUpdate
return dataPerUpdate
else
return 0
```

Andrey1994: No, no need to change it will just plot 4 new packages as in your example and as it should I think. I will just run it and test

daniellasry: What you're saying was true when the buffer had it's own thread, but it doesn't anymore.

Let's say FPs is 60. Let's say FTDI is so choopy it only updates every second. Initially, buffer will be empty. Then you will get 250 samples all at once. Buffer will be filled with 250 samples.

When we call getDataCount(), it will return 8, until the buffer is empty. getDataCount() is called every frame. So we will "consume" 8 samples every times it is called (60 times per second). At 8 samples per frame, it will take ~30 frames to consume the buffer. So the buffer will be consumed in half a second.

What am I missing?

Andrey1994: Yep, you are right I was thinking about it like its in another thread... And it was one more reason for threading its just easier

daniellasry: This isn't hard either:

deltaTimeSeconds = currentTime - timeOfLastCall
numSamples = deltaTimeSeconds * sampleRate

Andrey1994:
fixed and added interface to call control it in runtime. Didnt test switching during streaming(because I can not).  To test temporaty call it in initialize, need to rm it later

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -83,6 +83,73 @@ class BoardCytonSerialDaisy extends BoardCyton {
...
+    @Override
+    public boolean initializeInternal() {
+        boolean res = super.initializeInternal();
+        if (res) {
+            setSmoothingActive(true);
+        }
+        return res;
+    }
+
+    // synchronized is important to ensure that we dont free buffers during getting data
```
daniellasry, line 107: good thinking, since mouse clicks are on a different thread.

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -83,6 +83,73 @@ class BoardCytonSerialDaisy extends BoardCyton {
...
+    private Buffer<double[]> buffer = null;
+    private volatile boolean smoothData;
+
+    public BoardCytonSerialBase() {
+        super();
+        smoothData = false;
+    }
+
+    // todo remove it and call setSmoothingActive from UI button,
+    // remove this method from here completely
```
daniellasry, line 97: @retiutut in your change to add the button, remove this function

OpenBCI_GUI/Buffer.pde
```diff
@@ -0,0 +1,46 @@
...
+
+    Buffer(int samplingRate) {
+        // max delay smth like 2 seconds
+        this(samplingRate, samplingRate * 2);
+    }
+
+    public void addNewEntry(T object) {
+        while (this.size() >= maxSize) {
+            this.popFirstEntry();
+        }
```
daniellasry, line 26: IMO this is not ideal because we lose data if this happens. 
Instead this can be dealt with in getDataCount. If the buffer is over max by 10 samples, add 10 to "numSamples" before returning.

OpenBCI_GUI/Buffer.pde
```diff
@@ -0,0 +1,46 @@
...
+        this.add(object);
+    }
+
+    public T popFirstEntry() {
+        return this.poll();
+    }
+
+    public int getDataCount() {
+        Date date = new Date();
+        long currentTime = date.getTime();
```
daniellasry, line 36: I'm nitpicking, but if you want to get ms directly you can also use millis()

Andrey1994:
moved size check to getDataCount

OpenBCI_GUI/BoardCyton.pde
retiutut: @789 Are you sure? Should smoothing be off by default?

daniellasry: Yes, off by default (conor is cool with this)

PR #790, opened 2020-06-01
Cyton Accel has no off command

retiutut:
To account for this, hide the accelerometer on/off button for the off case.


daniellasry:
Wait, hang on. On ganglion you *can* turn off accelerometer. It's only on cyton that you can't. We still need to show this button for ganglion.

daniellasry:
Also, the same applied for Analog and Digital. On cyton you can't turn them off you can only switch to another mode.

daniellasry:
An alternate idea for a simple fix (it's a little weird):

- make it so that turning off accel switches to analog mode
- turning off analog switches to accel
- turning off digital switches to accel

retiutut:
>An alternate idea for a simple fix (it's a little weird):

make it so that turning off accel switches to analog mode
turning off analog switches to accel
turning off digital switches to accel

It kind of does this already. Sending team a video of this.

daniellasry:
> > An alternate idea for a simple fix (it's a little weird):
> 
> make it so that turning off accel switches to analog mode
> turning off analog switches to accel
> turning off digital switches to accel
> 
> It kind of does this already. Sending team a video of this.

You're right. There's just on missing piece in BoardCyton.pde:

replace:
```
    @Override
    public void setAccelerometerActive(boolean active) {
        if(active) {
            setBoardMode(CytonBoardMode.DEFAULT);
        }
        // no way of turning off accel.
    }
```

with:
```
    @Override
    public void setAccelerometerActive(boolean active) {
        if(active) {
            setBoardMode(CytonBoardMode.DEFAULT);
        }
        else {
            setBoardMode(CytonBoardMode.ANALOG);
        }
    }
```


That's all that's needed to fix this

retiutut:
Done

OpenBCI_GUI/W_Accelerometer.pde
```diff
@@ -216,8 +206,10 @@ class W_Accelerometer extends Widget {
             if (!accelBoard.isAccelerometerActive()) {
                 accelBoard.setAccelerometerActive(true);
                 output("Starting to read accelerometer");
+                accelModeButton.setString("Turn Accel. Off");
             } else {                    
                 accelBoard.setAccelerometerActive(false);
+                accelModeButton.setString("Turn Accel. On");
```
daniellasry, line 212: You should revert this to the way it was before (changing the button string on draw() or update())
Let the board class be the authority on the board state.
What if the board's state changes externally, without pressing the button?
For example, pressing "Turn Off Analog" will turn on Accel.

OpenBCI_GUI/W_Accelerometer.pde
```diff
@@ -158,18 +150,16 @@ class W_Accelerometer extends Widget {
...
         if (accelBoard.isAccelerometerActive()) {
             drawAccValues();
             draw3DGraph();
             accelerometerBar.draw();
+
+            if (currentBoard instanceof BoardGanglion) {
+                accelModeButton.draw();
+            }
+        } else {
+            accelModeButton.draw();
```
daniellasry, line 162: Shouldn't we just draw the button always?

retiutut: The button is not used in playback or synthetic mode. I think this part is right.

daniellasry: No it's not correct. It's not drawing in cyton when accel is on (but i can still click on it)

retiutut: Ok I see it.

PR #791, opened 2020-06-03
When channel disabled, turn off bias and srb2

daniellasry:
When disabling a channel:

- remember previous buttons settings
- dis-include bias
- disconnect srb2
- grey-out/disable all settings buttons for that channel.

When re-enabling channel:

- restore previous button settings
- re-enable all buttons for that channel

Bonus:

- Make it so you can press the channel on/off buttons when the settings controller is open.


daniellasry:
Added brainflow 3.0.3 and merging

PR #792, opened 2020-06-03
Fix autodropdown

daniellasry:
Fix the dropdowns auto-opening when you hover.

Dropdown opens when you click. 
Dropdown auto-closes when mouse leaves the area (as before).

removed "closeAllDropdowns" It was changing a boolean that was not accessed anywhere, so it was actually doing nothing.

Remove all rects drawing behind dropdowns (we can just use .setBackgroundColor())
Handle locking of other UI elements using events.


retiutut:
I've just reviewed the changes. 

Just to clarify, does this change effect literally every dropdown? Widget selectors? Widget settings dropdowns? Anything in the Control Panel?

daniellasry:
Only the widget selectors and widget settings dropdown were auto-opening. I removed the functionality from those.

but yes, all dropdowns are affected. I apply the makeDropdownsBetter() function to every scrollable list in the code.

Even the new NovaXR dropdowns will auto-close now.

OpenBCI_GUI/Interactivity.pde
```diff
@@ -294,12 +294,10 @@ synchronized void mouseReleased() {
 void makeScrollableListBetter(ScrollableList scrollList) {
     // there's a bug in control p5 where clicking on the scroll list does not
     // open it if you move the mouse while clicking. This fixes that.
-    scrollList.onRelease(new CallbackListener() {
+    scrollList.onEndDrag(new CallbackListener() {
```
retiutut, line 297: Nice!!!!

retiutut: It works!

retiutut:
Yes, ready for this change. Feels "right" when using the GUI.

PR #794, opened 2020-06-04
Fix extreme slowness when selecting Cyton Serial data source

daniellasry:
SerialPort.getCommPorts() is very expensive, and was being run on the main thread on every update

- Avoid calling getCommPorts on main thread for ganglion
- Make cyton serial discovery just like ganglion and wifi:
   - Call it on a separate thread
   - Call it once when selecting the data source, do not call again until user presses button
- There was duplicate code between autoconnect and Cyton discovery. Make this code shared.

### DIFF
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -1462,7 +1462,7 @@ class BLEBox {
         
     }
 
-    private void refreshPortListGanglion() {
+    private void refreshGanglionBLEList() {
         output("BLE Devices Refreshing");
         bleList.items.clear();
         
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -1468,6 +1468,8 @@ class BLEBox {
         
         Thread thread = new Thread(){
             public void run(){
+                refreshBLE.setString("SEARCHING...");
+                bleIsRefreshing = true;
                 final String comPort = getBLED112Port();
                 if (comPort != null) {
                     try {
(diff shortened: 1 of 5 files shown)


OpenBCI_GUI/ControlPanel.pde
```diff
@@ -1462,34 +1457,36 @@ class BLEBox {
...
-            Thread thread = new Thread(){
-                public void run(){
-                    refreshBLE.setString("SEARCHING...");
-                    bleIsRefreshing = true;
+        
+        Thread thread = new Thread(){
+            public void run(){
+                refreshBLE.setString("SEARCHING...");
+                bleIsRefreshing = true;
+                final String comPort = getBLED112Port();
```
daniellasry, line 1468: I moved the call to getBLED112Port() inside the thread.

daniellasry:
Turn out the reason getCommPorts was so slow for me, is because I set my device latency timer to 255 and forgot to change it back to 1 (FTDI buffer size). 

But any users who do not have the latency set to 1 (e.g. fan li's laptop), or who have another device plugged in with a slow response time might see the same issue .

daniellasry:
> Nice! It works now! I also check my latency is 16 ms. But I still have some issues in Time series widget. The plots is not smooth as it's supposed to be. I don't know if you have the same issue?

Yeah, I think this is normal if your latency is 16 ms. Have you tried setting it to 1?

Fan1117:
> > Nice! It works now! I also check my latency is 16 ms. But I still have some issues in Time series widget. The plots is not smooth as it's supposed to be. I don't know if you have the same issue?
> 
> Yeah, I think this is normal if your latency is 16 ms. Have you tried setting it to 1?

Oh I see. It works nicely when I change it to 1 ms! That's cool! Thanks!

PR #737, opened 2020-04-28
fix bit ordering in endian changes

wherrera10:
Bug fix: Bits are not swapped, just bytes -- please test this on Apple hardware before merging it if you can, I'm limited in how I can test it.  Write the bdf file and check that it looks ok in EDFBROWSER, for example.

PR #800, opened 2020-06-06
wip sdcard board

Andrey1994:

Any ideas why I see arrayindexoutofrange and changes in UI are welcome


Andrey1994:
fixed index error, need to change smth in control panel to set new data source

Andrey1994:
Maybe intead Integer.parseInt we need to use this https://github.com/brainflow-dev/brainflow/blob/master/src/utils/inc/custom_cast.h#L7 (in java of course) but old SDConverter didnt...

OpenBCI_GUI/FileBoard.pde
```diff
@@ -0,0 +1,12 @@
+interface FileBoard {
```
daniellasry, line 1: Is the plan to have DataSourcePlayback also implement the FileBoard interface?

OpenBCI_GUI/DataSourceSDCard.pde
```diff
@@ -0,0 +1,264 @@
...
+    private int samplingRate;
+    private ArrayList<double[]> data;
+    private int[] exgChannels;
+    private int totalChannels;
+    private boolean streaming;
+    private double startTime;
+    private int counter;
+    private int currentSample;
+    private int timeOfLastUpdateMS;
+    private boolean initialized;
```
daniellasry, line 13: is this boolean used for anything?

OpenBCI_GUI/DataSourceSDCard.pde
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+
+    private void parseRow(String[] row, int numChannels) {
+        double[] line = new double[totalChannels];
+        if (row.length < numChannels - 1) {
+            return;
+        }
+        for (int i = 0; i < numChannels; i++) {
+            Integer val = Integer.parseInt(row[i], 16);
```
daniellasry, line 80: This looks fine. It should work.

OpenBCI_GUI/DataSourceSDCard.pde
```diff
@@ -0,0 +1,264 @@
...
+        if (row.length >= numChannels + 4) {
+            accel_x = Integer.parseInt(row[numChannels + 1], 16);
+            accel_y = Integer.parseInt(row[numChannels + 2], 16);
+            accel_z = Integer.parseInt(row[numChannels +3], 16);
+        }
+        line[line.length - 4] = BoardCytonConstants.accelScale * accel_x;
+        line[line.length - 3] = BoardCytonConstants.accelScale * accel_y;
+        line[line.length - 2] = BoardCytonConstants.accelScale * accel_z;
+        // add timestamp
+        double delay = (numChannels == 16) ? (1.0 / 125) : (1.0 / 250);
```
daniellasry, line 97: you could use getSampleRate() here

daniellasry:
I think it looks good so far

daniellasry:
I did a test and indeed it seems liek there is a problem with parsing the data. I tested square wave, and it looks exactly like the issue we had with BDF: https://github.com/OpenBCI/OpenBCI_GUI/pull/737/files

daniellasry:
Try these parsing functions:

```
    private int parseInt24Hex(String hex) {
        if (hex.charAt(0) > '7') {  // if the number is negative	
            hex = "FF" + hex;   // keep it negative	
        } else {                  // if the number is positive	
            hex = "00" + hex;   // keep it positive	
        }

        return unhex(hex);
    }

    private int parseInt16Hex(String hex) {
        if (hex.charAt(0) > '7') {  // if the number is negative	
            hex = "FFFF" + hex;   // keep it negative	
        } else {                  // if the number is positive	
            hex = "0000" + hex;   // keep it positive	
        }

        return unhex(hex);
    }
```

OpenBCI_GUI/FileBoard.pde
Andrey1994: I think it does, if not - need to add it

OpenBCI_GUI/DataSourceSDCard.pde
Andrey1994: Probably it was but now not, can remove

OpenBCI_GUI/DataSourceSDCard.pde
Andrey1994: Before for the first sample getSampleRate was set after first parsed package, so I hardcoded it. Now its possible

Andrey1994:
What is unhex? There is no such function in java and I dont see it in gui

Andrey1994:
And it looks strange... its not a conversion from int24/16 I use in brainflow....

OpenBCI_GUI/FileBoard.pde
Andrey1994: fixed

daniellasry:
unhex is from processing: https://processing.org/reference/unhex_.html

I don't know why it works and parseInt does not

daniellasry:
Could you change the string "CONVERT SD FOR PLAYBACK" to something like "SELECT SD FOR PLAYBACK"?

Also, could you add a line in CHANGELOG.md, under Improvements, saying that SD files can be played back directly without converting?

Andrey1994:
Its already renamed: https://github.com/OpenBCI/OpenBCI_GUI/pull/800/files#diff-5295e497df226f2138eae5e540267aeaR1066 if there are other places I dont see them

daniellasry:
Yes, ControlPanel.pde line 2549

daniellasry:
Also, WidgetManager.pde line 69, replace that with if(currentBoard instanceof DataSourcePlayback)
We don't want that widget showing in SD card playback.

Andrey1994:
> Also, WidgetManager.pde line 69, replace that with if(currentBoard instanceof DataSourcePlayback)
> We don't want that widget showing in SD card playback.

Why not?

Andrey1994:
Ah, makes sense cause probably this widget set playback file

daniellasry:
One bug: If i select playback file then select SD file, it will run playback file
Maybe you need to clear the playback file name when user selects sd card?

PR #809, opened 2020-06-16
Brainflow filters

Andrey1994:
Use brainflow filters instead matlab coeffs.

My original idea was to add an abstract class or interface and represent each filter as an object with method like
```
float[] transformData(float[] inputData);
```
and store all these objects in an ordered set(need to implement Comparable interface also) and call these methods in a chain. So it would be smth like simple autogenerated node graph...

But I decided to keep everything simple for now, and just replace old filters by brainflow filters
I tested it only with synthetic data, so it will be great if we all test it with real data. There is a chance that smth is wrong

@retiutut I temporary removed smth from software settings, could you restore it using new approach?

### DIFF
--- a/OpenBCI_GUI/DataProcessing.pde
+++ b/OpenBCI_GUI/DataProcessing.pde
@@ -93,7 +93,7 @@ class DataProcessing {
     float data_std_uV[];
     float polarity[];
     boolean newDataToSend;
-    BandPassRanges bpRange = BandPassRanges.OneToFifty;
+    BandPassRanges bpRange = BandPassRanges.FiveToFifty;
     BandStopRanges bsRange = BandStopRanges.Sixty;
     final int[] processing_band_low_Hz = {
         1, 4, 8, 13, 30
--- a/OpenBCI_GUI/DataProcessing.pde
+++ b/OpenBCI_GUI/DataProcessing.pde
@@ -4,6 +4,9 @@
 //------------------------------------------------------------------------
 import ddf.minim.analysis.*; //for FFT
 
+import brainflow.DataFilter;
+import brainflow.FilterTypes;
+
 DataProcessing dataProcessing;
 String curTimestamp;
 HashMap<Integer,String> index_of_times;
@@ -109,15 +112,11 @@ void initializeFFTObjects(FFT[] fftBuff, float[][] dataProcessingRawBuffer, int 
 class DataProcessing {
     private float fs_Hz;  //sample rate
     private int nchan;
-    final int N_FILT_CONFIGS = 5;
-    FilterConstants[] filtCoeff_bp = new FilterConstants[N_FILT_CONFIGS];
-    final int N_NOTCH_CONFIGS = 3;
-    FilterConstants[] filtCoeff_notch = new FilterConstants[N_NOTCH_CONFIGS];
-    private int currentFilt_ind = 3;
-    private int currentNotch_ind = 0;  // set to 0 to default to 60Hz, set to 1 to default to 50Hz
     float data_std_uV[];
     float polarity[];
     boolean newDataToSend;
+    BandPassRanges bpRange = BandPassRanges.OneToFifty;;
+    BandStopRanges bsRange = BandStopRanges.Sixty;
     final int[] processing_band_low_Hz = {
         1, 4, 8, 13, 30
     }; //lower bound for each frequency band of interest (2D classifier only)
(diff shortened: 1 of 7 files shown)


Andrey1994:
Btw I would like to add brainflow denoising as one more option for signal processing

Andrey1994:
Tested with real data, notch filter works fine for me. Dont know how to validate bandpass

retiutut:
>@retiutut I temporary removed smth from software settings, could you restore it using new approach?

That's fine and it makes sense. New UI will be much different since there will be more options.

Andrey1994:
> > @retiutut I temporary removed smth from software settings, could you restore it using new approach?
> 
> That's fine and it makes sense. New UI will be much different since there will be more options.

I wanna merge as is for now and update UI and add more options later on, so will be nice to restore it for now

OpenBCI_GUI/DataProcessing.pde
```diff
@@ -87,15 +90,11 @@ void initializeFFTObjects(FFT[] fftBuff, float[][] dataProcessingRawBuffer, int
...
-    final int N_FILT_CONFIGS = 5;
-    FilterConstants[] filtCoeff_bp = new FilterConstants[N_FILT_CONFIGS];
-    final int N_NOTCH_CONFIGS = 3;
-    FilterConstants[] filtCoeff_notch = new FilterConstants[N_NOTCH_CONFIGS];
-    private int currentFilt_ind = 3;
-    private int currentNotch_ind = 0;  // set to 0 to default to 60Hz, set to 1 to default to 50Hz
     float data_std_uV[];
     float polarity[];
     boolean newDataToSend;
+    BandPassRanges bpRange = BandPassRanges.OneToFifty;;
```
retiutut, line 96: double semi colon here `;;`

OpenBCI_GUI/DataProcessing.pde
```diff
@@ -113,289 +112,45 @@ class DataProcessing {
...
-
-                //create the bandpass filter
-                filtCoeff_bp[Ifilt] =  new FilterConstants(b, a, filt_txt, short_txt);
-        } //end loop over band pass filters
     }
-    //end defineFilters method
 
     public String getFilterDescription() {
-        return filtCoeff_bp[currentFilt_ind].name + ", " + filtCoeff_notch[currentNotch_ind].name;
+        return bpRange.getDescr();
```
retiutut, line 118: This seems Ok for now. 👍

OpenBCI_GUI/FilterEnums.pde
```diff
@@ -0,0 +1,70 @@
...
+
+    public String getDescr() {
+        if (freq == null) {
+            return "None";
+        }
+        return freq.toString();
+    }
+}
+
+public enum BandPassRanges
```
retiutut, line 32: I like this! 👍

OpenBCI_GUI/SoftwareSettings.pde
```diff
@@ -787,13 +787,6 @@ class SoftwareSettings {
 
         //}//end case for all objects in JSON
 
-        //Apply notch
```
retiutut, line 790: 👍

OpenBCI_GUI/DataProcessing.pde
```diff
@@ -113,289 +112,45 @@ class DataProcessing {
...
-    void processChannel(int Ichan, float[][] data_forDisplay_uV, float[] prevFFTdata) {            
+    private synchronized void processChannel(int Ichan, float[][] data_forDisplay_uV, float[] prevFFTdata) {            
         int Nfft = getNfftSafe();
         double foo;
 
         //filter the data in the time domain
-        filterIIR(filtCoeff_notch[currentNotch_ind].b, filtCoeff_notch[currentNotch_ind].a, data_forDisplay_uV[Ichan]); //notch
-        filterIIR(filtCoeff_bp[currentFilt_ind].b, filtCoeff_bp[currentFilt_ind].a, data_forDisplay_uV[Ichan]); //bandpass
+        try {
+            double[] tempArray = floatToDoubleArray(data_forDisplay_uV[Ichan]);
```
daniellasry, line 141: let's make a //todo note for later to convert data processing to use only double (it will be easier now since we are using brainflow filters).

Then we will only cast to floats once for drawing purposes.

OpenBCI_GUI/FilterEnums.pde
```diff
@@ -0,0 +1,70 @@
...
+
+    public BandStopRanges next()
+    {
+        return vals[(this.ordinal() + 1) % vals.length];
+    }
+
+    public String getDescr() {
+        if (freq == null) {
+            return "None";
+        }
```
daniellasry, line 27: Is this necessary? wouldn't freq.toString(); return "None" anyway?

OpenBCI_GUI/FilterEnums.pde
```diff
@@ -0,0 +1,70 @@
...
+
+    public BandPassRanges next()
+    {
+        return vals[(this.ordinal() + 1) % vals.length];
+    }
+
+    public String getDescr() {
+        if ((start == null) || (stop == null)) {
+            return "None";
+        }
```
daniellasry, line 67: same comment as above

OpenBCI_GUI/FilterEnums.pde
daniellasry: Ignore this comment, I was confused.

OpenBCI_GUI/DataProcessing.pde
Andrey1994: Yep I thought about it but didnt implement in this PR

Andrey1994:
No problem, lets test first

daniellasry:
I finally had a chance to compare old vs new. I used the same playback file with a sqaure wave.

Surprisingly, all filters look *exactly* the same. Absolutely identical between old and new.

The only differences were in the UI:
- The default BP filter is 5-50 in old GUI. In this PR it's 1-50
- The buttons display integers and "Hz" in old GUI. Old vs new: 
image
image 

These should be easy fixes you can make in the enum

daniellasry:
I also checked for any performance hit... no difference.

Andrey1994:
I would not test filters with square waves but its good news

Andrey1994:
> I also checked for any performance hit... no difference.

In this case it can be faster wo conversion to\from double

PR #810, opened 2020-06-18
add streaming board

Andrey1994:

WIP: add streaming board.

TODOs:

* add UI elements(need help)
* I dont know good solution for other interfaces(Accel,PPG,....) which master board may or may not implement. Add only EXG data for now and probably its good enough

Instructions to run:

Increase runtime in brainflow_get_data from 10sec to smth bigger and run
```
python brainflow_get_data.py --log --board-id -1 --streamer-params streaming_board://225.1.1.1:6677
```

### DIFF
--- a/OpenBCI_GUI/BoardBrainFlowStreaming.pde
+++ b/OpenBCI_GUI/BoardBrainFlowStreaming.pde
@@ -1,0 +1,69 @@
+import brainflow.*;
+
+class BoardBrainFlowStreaming extends BoardBrainFlow {
+
+    private int masterBoardId;
+    private String ipAddress;
+    private int ipPort;
+
+    public BoardBrainFlowStreaming(int masterBoardId, String ipAddress, int ipPort) {
+        super();
+        this.masterBoardId = masterBoardId;
+        this.ipAddress = ipAddress;
+        this.ipPort = ipPort;
+    }
+
+    // implement mandatory abstract functions
+    @Override
+    protected BrainFlowInputParams getParams() {
+        BrainFlowInputParams params = new BrainFlowInputParams();
+        params.ip_address = ipAddress;
+        params.ip_port = ipPort;
+        params.other_info = Integer.toString(masterBoardId, 10);
+        return params;
+    }
+
+    // for streaming board need to use master board id in function like  get_eeg_channels
+    @Override
+    public BoardIds getBoardId() {
+        return BoardIds.from_code(masterBoardId);
+    }
+
+    @Override
+    public boolean initializeInternal() {
+        try {
+            // here we need to provide board id of streaming board
+            boardShim = new BoardShim (BoardIds.STREAMING_BOARD.get_code(), getParams());
+            try {
+                BoardShim.enable_dev_board_logger();
+                BoardShim.set_log_file("brainflow_log.txt");
+            } catch (BrainFlowError e) {
+                e.printStackTrace();
+            }
+            boardShim.prepare_session();
+            return true; 
+
+        } catch (Exception e) {
+            boardShim = null;
+            outputError("ERROR: " + e + " when initializing Brainflow board. Data will not stream.");
+            e.printStackTrace();
+            return false;
+        }
+    }
+
+    @Override
+    public void setEXGChannelActive(int channelIndex, boolean active) {
+        // do nothing here
+    }
+
+    @Override
+    public boolean isEXGChannelActive(int channelIndex) {
+        return true;
+    }
+
+    @Override
+    protected void addChannelNamesInternal(String[] channelNames) {
+        // do nothing here
+    }
+
+};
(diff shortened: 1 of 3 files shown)


daniellasry:
For UI richard or I can help.

For the interfaces, probably the thing to do is to have it implement all interfaces, like playback data source.

In retrospect, maybe using interfaces to build board functionality was not the best. Using composition would have been better because we could compose board functionality at runtime.

Andrey1994:
Added accel interface, I think for now we dont need to add other interfaces

OpenBCI_GUI/BoardBrainFlowStreaming.pde
```diff
@@ -0,0 +1,98 @@
...
+            boardShim.prepare_session();
+            return true; 
+
+        } catch (Exception e) {
+            boardShim = null;
+            outputError("ERROR: " + e + " when initializing Brainflow board. Data will not stream.");
+            e.printStackTrace();
+            return false;
+        }
+    }
```
daniellasry, line 52: Damn, it's unfortunate that this code had to be copied, but I understand why.

I have an idea to fix, although it's a little weird:
Have getBoardId() return BoardIds.STREAMING_BOARD, but after initialization, it returns masterBoardId.

```
    public BoardIds getBoardId() {
        if (boardShim == null) {
            return BoardIds.STREAMING_BOARD
        }
        return BoardIds.from_code(masterBoardId);
    }
```

OpenBCI_GUI/BoardBrainFlowStreaming.pde
```diff
@@ -0,0 +1,98 @@
+import brainflow.*;
+
+class BoardBrainFlowStreaming extends BoardBrainFlow implements AccelerometerCapableBoard {
+
+    private int masterBoardId;
```
daniellasry, line 5: let's use the BoardIds enum type instead of int (same in the constructor arg)

OpenBCI_GUI/BoardBrainflow.pde
```diff
@@ -22,14 +22,11 @@ abstract class BoardBrainFlow extends Board {
...
     public boolean initializeInternal() {
-        // initiate the board shim
         try {
-
             boardShim = new BoardShim (getBoardIdInt(), getParams());
-            // for some reason logger configuration doesnt work in contructor or static initializer block
-            // and it looks like smth processing specific
             try {
                 BoardShim.enable_dev_board_logger();
+                BoardShim.set_log_file("brainflow_log.txt");
```
daniellasry, line 29: nice!

Can we set this to the same location as the console log files? 
```
BoardShim.set_log_file(settings.consoleDataPath+"Brainflow_"+DirectoryManager.getFileNameDateTime()+".txt");
```

will place the file in C:\Users\user\Documents\OpenBCI_GUI\Console_Data

Andrey1994:
Just to remind - temp code to test it wo UI still exists, you approved it too soon)

daniellasry:
I think you can comment out the test code before sumitting and we will fix it up when we add a UI

Andrey1994:
We can add UI directly to this branch(its not in my fork)

daniellasry:
Sounds like a plan.

OpenBCI_GUI/BoardBrainFlowStreaming.pde
Andrey1994: I think its better to copy one method, it looks really weird....
Also if this method will be implemented like you wrote it will be impossible to query board info(sampling rate, eeg channels,....) before initialization. Right now I think we dont have such cases but maybe will add later

Andrey1994: or maybe we even have such cases now, as I remember getEXGChannels can be called before initialize

OpenBCI_GUI/BoardBrainflow.pde
Andrey1994: done

OpenBCI_GUI/BoardBrainFlowStreaming.pde
Andrey1994: done

OpenBCI_GUI/BoardBrainFlowStreaming.pde
daniellasry: you make good points

retiutut:
@Andrey1994 Can you start a list here for the UI requirements? What should be included?

Andrey1994:
> @Andrey1994 Can you start a list here for the UI requirements? What should be included?

Just these three values:
```
currentBoard = new BoardBrainFlowStreaming(BoardIds.SYNTHETIC_BOARD, "225.1.1.1", 6677);
```
And new eeg_data_source of course, smth like BrainFlow_Streaming_board

Andrey1994:
Not related to this PR but BrainFlow streaming params(text field) can also be added for other BrainFlow boards, somewhere in Cyton/Ganglion/Synthetic/etc config boxes

retiutut:
@Andrey1994 UI has been added, and I have tested using Ganglion! If users select wrong board type, I think BrainFlow will throw an error on prepare_session.

PR #812, opened 2020-06-23
Cleanly stop streaming and close board on quit

daniellasry:
Simply call haltSystem() on quit.

### DIFF
--- a/OpenBCI_GUI/BoardNull.pde
+++ b/OpenBCI_GUI/BoardNull.pde
@@ -21,12 +21,12 @@ class BoardNull extends Board {
 
     @Override
     public void startStreaming() {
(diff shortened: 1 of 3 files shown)


Andrey1994:
What is haltSystem? I dont see definition in code and at processing.org

daniellasry:
It's defined in OpenBCI_GUI.pde

daniellasry:
It's the function that gets called when you stop a session

daniellasry:
offline discussion: andrey pointed out that brainflow should already be stopping the stream when the board object is destructed. It seems like it does.

Still, i think this change might be for the best, since it covers things like closing the SD card, closing the log file on exit... It will avoid more problems like that in the future.

PR #813, opened 2020-06-23
Fix autoconnect issue (conor reported)

daniellasry:
When disconnecting and reconnecting the dongle, sometimes the name will appear as VCP0. Added "VCP0" to the list of possible names.

Repro steps for conor's bug:
- Autoconnect to cyton serial
- stop session
- unplug dongle
- plug dongle
- press autoconnect

### DIFF
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -1360,6 +1360,12 @@ class ComPortBox {
                 initButtonPressed();
                 buttonHelpText.setVisible(false);
             }
+            else {                
+                outputWarn("Found a Cyton dongle, but could not connect to the board.");
+            }
+        }
+        else {
+            outputWarn("No Cyton dongles were found.");
         }
     }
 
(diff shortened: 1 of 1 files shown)


retiutut:
@daniellasry Did you test this on Mac? Was this issue happening on Mac and Windows?

daniellasry:
I did not test it on mac, but it would be a good idea

retiutut:
Not able to reproduce on my iMac. I'll try on Windows 10 tomorrow/today.

daniellasry:
On mac, I can't reproduce it either, with or without this fix.
I think you would need multiple cyton boards to reproduce it on mac. On windows we are able to reproduce with one board.

I think we should merge this and ask conor if it's fixed. At any rate, this fixes a legit issue on windows. Hopefully it fixes the mac issue conor was having. He can let us know.

@retiutut sounds good to merge this now?

retiutut:
I would still like to reproduce also on Windows, I think I got caught up doing support for Cyton.

PR #815, opened 2020-06-24
cleanup code when resizing gui

daniellasry:
- Remove unnecessary code that stops drawing when resizing
- remove annoying printouts when resizing GUI which were also slowing things down.

### DIFF
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -326,7 +326,6 @@ void delayedSetup() {
     frame.addComponentListener(new ComponentAdapter() {
         public void componentResized(ComponentEvent e) {
             if (e.getSource()==frame) {
-                println("OpenBCI_GUI: setup: RESIZED");
                 settings.screenHasBeenResized = true;
                 settings.timeOfLastScreenResize = millis();
                 // initializeGUI();
@@ -792,12 +791,9 @@ void systemUpdate() { // for updating data values and variables
     }
     if (systemMode == SYSTEMMODE_POSTINIT) {
         processNewData();
-
-        // gui.cc.update(); //update Channel Controller even when not updating certain parts of the GUI... (this is a bit messy...)
-
+        
         //alternative component listener function (line 177 - 187 frame.addComponentListener) for processing 3,
         if (settings.widthOfLastScreen != width || settings.heightOfLastScreen != height) {
-            println("OpenBCI_GUI: setup: RESIZED");
             settings.screenHasBeenResized = true;
             settings.timeOfLastScreenResize = millis();
             settings.widthOfLastScreen = width;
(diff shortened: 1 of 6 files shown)


OpenBCI_GUI/W_Networking.pde
```diff
@@ -717,7 +717,6 @@ class W_Networking extends Widget {
         itemWidth = int(map(width, 1024, 1920, 100, 120)) - 4;
         
         int dropdownsItemsToShow = int((this.h0 * datatypeDropdownScaling) / (this.navH - 4));
-        println("Networking Data Types || show num dropdowns = " + dropdownsItemsToShow);
```
retiutut, line 720: Yea this println can go.

OpenBCI_GUI/Widget.pde
```diff
@@ -224,7 +224,6 @@ class Widget{
 
     private void resizeWidgetSelector() {
         int dropdownsItemsToShow = int((h0 * widgetDropdownScaling) / (navH - 4));
-        //println("Widget " + widgetTitle +  " || show num dropdowns = " + dropdownsItemsToShow);
```
retiutut, line 227: I guess this can go too, since it works.

OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -860,22 +851,7 @@ void systemDraw() { //for drawing to the screen
...
-        if ((millis() - settings.timeOfGUIreinitialize) > settings.reinitializeGUIdelay) {
-            // println("attempting to draw GUI...");
-            try {
-                // println("GUI DRAW!!! " + millis());
-                //draw GUI widgets (visible/invisible) using widget manager
-                wm.draw();
-            } catch (Exception e) {
-                println(e.getMessage());
-                settings.reinitializeGUIdelay = settings.reinitializeGUIdelay * 2;
-                println("OpenBCI_GUI: systemDraw: New GUI reinitialize delay = " + settings.reinitializeGUIdelay);
```
retiutut, line 873: I don't think I've ever seen this exception, and hard to tell from the code why we can't just try to draw the GUI.

retiutut:
```
DEFAULT]: Data stream started.
OpenBCI_GUI: openNewLogFile: opened ODF output file: /Users/Richard/Documents/OpenBCI_GUI/Recordings/OpenBCISession_2020-06-24_23-12-52/OpenBCI-RAW-2020-06-24_23-13-04.txt
handleDraw() called before finishing
SHUTDOWN HOOK
SoftwareSettings: 3 active widgets saved!
SoftwareSettings.pde:1229:0:1229:0: GLException: Caught ThreadDeath: null on thread main-FPSAWTAnimator#00-Timer0-FPSAWTAnimator#00-Timer1-FPSAWTAnimator#00-Timer2-FPSAWTAnimator#00-Timer3
```

Caused by 
```
void imposeMinimumGUIDimensions() {
    //impose minimum gui dimensions
    if (width < settings.minGUIWidth || height < settings.minGUIHeight) {
        if (width < settings.minGUIWidth) win_x = settings.minGUIWidth;
        if (height < settings.minGUIHeight) win_y = settings.minGUIHeight;
        surface.setSize(win_x, win_y); <--Here
    }
}
```

Which is kind of nice to have.

retiutut: On second test, unable to reproduce!

I can force the above error if I aggressively resize the GUI while running 16Ch Synthetic w/ 20 seconds of TimeSeries and Accelerometer open. (to stress frame rate)

retiutut: On third test with development branch, crash happens faster because of all these extra println()'s being removed in this PR.

Good changes!

PR #817, opened 2020-06-25
Smoothing buffer was filling up to the max

daniellasry:
Conor was right. The smoothing buffer is quickly filling up to its max size and introducing a 2 second delay.

The fix is to use ceil() when calculating the number of samples to deliver, and err on the side of delivering more samples.

### DIFF
--- a/OpenBCI_GUI/Buffer.pde
+++ b/OpenBCI_GUI/Buffer.pde
@@ -5,17 +5,17 @@ public class Buffer<T> extends LinkedList<T> {
 
     private int samplingRate;
     private int maxSize;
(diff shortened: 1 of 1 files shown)


daniellasry:
> I dont fully understand how ceil fixes it but I am ok with this change

```
numSamples = (int)(samplingRate * deltaTimeSeconds);
```

This is imprecise, because we are truncating the float result into int. 

Let's say that the calculation thinks that we should release 16.6 samples this frame. We truncate that to 16. So we don't release enough samples to keep up with the sample rate. By using ceil we release 17 frames.

PR #820, opened 2020-06-27
NovaXR Aux Widget 2.0

retiutut:
Requirements:

2 Channels PPG
1 Channel EDA
1 Channel / Display Battery Info

- [x] fix novaAux autoscale to show only values greater than or equal to 0

### DIFF
--- a/OpenBCI_GUI/BatteryInfoCapableBoard.pde
+++ b/OpenBCI_GUI/BatteryInfoCapableBoard.pde
@@ -1,5 +1,5 @@
 
 interface BatteryInfoCapableBoard {
 
-    public Integer getBatteryInfo();
+    public Integer getBatteryChannel();
 };
--- a/OpenBCI_GUI/BatteryInfoCapableBoard.pde
+++ b/OpenBCI_GUI/BatteryInfoCapableBoard.pde
@@ -1,0 +1,5 @@
+
+interface BatteryInfoCapableBoard {
+
+    public Integer getBatteryInfo();
+};
--- a/OpenBCI_GUI/W_NovaAux.pde
+++ b/OpenBCI_GUI/W_NovaAux.pde
@@ -284,7 +284,7 @@ class AuxReadBar{
 
         if (auxValuesPosition == 1 || auxValuesPosition == 2) {
             channels = ppgBoard.getPPGChannels(); 
-            //channelNumber = auxValuesPosition - 1;
+            channelNumber = auxValuesPosition - 1;
         } else if (auxValuesPosition == 3) {
             channels = edaBoard.getEDAChannels(); 
         } else {
--- a/OpenBCI_GUI/W_NovaAux.pde
+++ b/OpenBCI_GUI/W_NovaAux.pde
@@ -13,29 +13,26 @@ class W_NovaAux extends Widget {
 
     private AuxReadBar[] analogReadBars;
 
-    private int[] xLimOptions = {0, 1, 3, 5, 10, 20}; // number of seconds (x axis of graph)
-    private int[] yLimOptions = {0, 50, 100, 200, 400, 1000, 10000}; // 0 = Autoscale ... everything else is uV
-
+    public int[] xLimOptions = {1, 3, 5, 10, 20}; // number of seconds (x axis of graph)
+    public int[] yLimOptions = {0, 50, 100, 200, 400, 1000, 10000}; // 0 = Autoscale ... everything else is uV
+    //Used to set text in dropdown menus when loading Analog Read settings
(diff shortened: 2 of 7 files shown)


OpenBCI_GUI/W_NovaAux.pde
```diff
@@ -284,12 +292,16 @@ class AuxReadBar{
...
 
-        if (auxValuesPosition == 1) {
+        if (auxValuesPosition == 1 || auxValuesPosition == 2) {
+            channels = ppgBoard.getPPGChannels(); 
+            channelNumber = auxValuesPosition - 1;
+        } else if (auxValuesPosition == 3) {
             channels = edaBoard.getEDAChannels(); 
         } else {
-            channels = ppgBoard.getPPGChannels(); 
+            channels[0] = batteryBoard.getBatteryInfo();
```
retiutut, line 304: Getting NullPointerException here @Andrey1994

Andrey1994: Because there is a code
```
        edaBoard = (EDACapableBoard) currentBoard;
        ppgBoard = (PPGCapableBoard) currentBoard;
```
and you forgot to add the same for batteryboard

Andrey1994:
BTW I think unlike novaxr synthetic board has one PPG channel so to test it like in novaxr you can duplicate channel temporary

retiutut:
👍 

Screen Shot 2020-06-27 at 1 12 04 AM

Andrey1994:
I think for "window sync" window size should be 5 instead 20, or remove sync option completely

daniellasry:
To avoid getting the branches out of sync, ideally this should be merged into the development branch, then we can do a merge from development to novaxr-dev

You can temporarily change the novaXREnabled flag locally then change it back before submitting

Andrey1994:
> To avoid getting the branches out of sync, ideally this should be merged into the development branch, then we can do a merge from development to novaxr-dev
> 
> You can temporarily change the novaXREnabled flag locally then change it back before submitting

we will still be able to apply this change to development branch even if we merge it here first.
```
git format-patch -1
git checkout developement
git am %patch name%
git push -u origin development
```

But maybe its better to merge to development first

retiutut:
I had just merged development into novaxr-dev and started from that.

retiutut:
I'd like to make a nice looking battery meter soon, as requested, but I would like to confirm information on value ranges. Then, I can custom set the graph Y range for each Aux channel by type.

retiutut:
Screen Shot 2020-06-28 at 11 00 34 PM

OpenBCI_GUI/W_NovaAux.pde
```diff
@@ -284,17 +279,21 @@ class AuxReadBar{
...
```
daniellasry, line 291: We only have one channel per AuxReadBar, correct?
Let's get rid of the "channelNumber" variable. Instead of "channels" let's have "channel" be a single int. Code becomes:

```
int channel = 0;
switch (auxValuesPosition) {
    case 1:
        channel = ppgBoard.getPPGChannels()[0];
        break;
    case 2:
        channel = ppgBoard.getPPGChannels()[1];
        break;
    case 3:
        channel = edaBoard.getEDAChannels()[0];
        break;
    default:
        channel = batteryBoard.getBatteryChannel();
        break;
}
```

daniellasry: Also, this should be moved to the constructor. We only need to figure out our channel once... not evey update.

retiutut: I like it better this way.

OpenBCI_GUI/BatteryInfoCapableBoard.pde
```diff
@@ -0,0 +1,5 @@
+
+interface BatteryInfoCapableBoard {
+
+    public Integer getBatteryInfo();
```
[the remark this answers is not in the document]
retiutut: Done!

OpenBCI_GUI/W_NovaAux.pde
```diff
@@ -279,21 +279,25 @@ class AuxReadBar{
...
+                break;
+            case 2:
+                channel = ppgBoard.getPPGChannels()[1];
+                break;
+            case 3:
+                channel = edaBoard.getEDAChannels()[0];
+                break;
+            default:
+                channel = batteryBoard.getBatteryChannel();
+                break;
```
daniellasry, line 295: perfect, but could you move it to the constructor of W_NovaAux? The channel variable only needs to be calculated once and can be saved as a private var.

retiutut: Done!

retiutut:
Changed base branch.

PR #824, opened 2020-06-30
No Autoload SoftwareSettings

retiutut:
Allow all users to start all sessions with default settings. Users still have the choice to save/load custom settings, and last session's settings will still be saved on close. Default settings are still saved at the start of each session, which is the forward-compatible solution that allows an easy "reset" option.

Serious users may not like having to turn on expert mode every time, in order to load last session's settings using keyboard shortcuts. Might still be useful to autoload specific app-wide parameters such as this.

### DIFF
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -581,15 +581,15 @@ void initSystem() {
 
     verbosePrint("OpenBCI_GUI: initSystem: -- Init 4 -- " + millis());
 
-    //DISABLE SOFTWARE SETTINGS FOR NOVAXR
-    if (eegDataSource != DATASOURCE_NOVAXR) {
+    
+    if (eegDataSource != DATASOURCE_NOVAXR) { //don't save default settings for NovaXR
         if (!abandonInit) {
-            //Init software settings: create default settings files, load user settings, etc.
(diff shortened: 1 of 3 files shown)


OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -581,15 +581,15 @@ void initSystem() {
 
     verbosePrint("OpenBCI_GUI: initSystem: -- Init 4 -- " + millis());
 
-    //DISABLE SOFTWARE SETTINGS FOR NOVAXR
-    if (eegDataSource != DATASOURCE_NOVAXR) {
+    
+    if (eegDataSource != DATASOURCE_NOVAXR) { //don't save default settings for NovaXR
```
Andrey1994, line 585: I dont fully understand why we need special handling for novaxr, novaxr doesnt look like corner case in terms of software settings

daniellasry: I also wonder if this is still necessary. Maybe there was a bug before and we fixed it since?

retiutut: I just didn't take the time to make sure it works, and I was asked to do this.

I think only thing blocking functional softwareSettings for NovaXR is two NovaAux widget settings.

PR #825, opened 2020-06-30
use brainflow synthetic board

Andrey1994:

image

Its not exactly the same sin waves as in gui's synthetic board and for accel it returns data using normal distribution


retiutut:
Screen Shot 2020-07-02 at 2 53 55 PM

Would be better if the values could be intervals of 5 and 10 instead, to check exact value in FFT and Spectrogram Widgets.

PR #826, opened 2020-06-30
Gui v5 ui fixes

retiutut:
- don't autoload settings for all data types

Screen Shot 2020-06-30 at 3 31 11 PM

### DIFF
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -581,15 +581,15 @@ void initSystem() {
 
     verbosePrint("OpenBCI_GUI: initSystem: -- Init 4 -- " + millis());
 
-    //DISABLE SOFTWARE SETTINGS FOR NOVAXR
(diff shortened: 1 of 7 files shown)


OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -584,15 +584,15 @@ void initSystem() {
...
+    if (eegDataSource != DATASOURCE_NOVAXR) { //don't save default settings for NovaXR
         if (!abandonInit) {
-            //Init software settings: create default settings files, load user settings, etc.
+            //Init software settings: create default settings file that is datasource unique
             settings.init();
             settings.initCheckPointFive();
         } else {
             haltSystem();
-            outputError("Failed to connect. Check that the device is powered on and in range.");
+            outputError("Failed to save Default Settings during Init. Please submit an Issue on GitHub.");
```
retiutut, line 595: This would really only happen if we are unable to write files to the Documents folder.

PR #827, opened 2020-07-01
Packet loss tracking widget

daniellasry:
- Created a PacketLossTracker class
- Board class holds instance of PacketLossTracker and feeds it samples.
- PacketLossTracker detects packet loss and keeps track of it
- Create W_Packetloss that displays packet loss statistics

CytonDaisySerial and Gangion have weird packed indices.
- CytonDaisySerial: indices 0-254, even numbers only
- Ganglion with Accel: indices 0-100. All indices are duplicated except zero. (0, 1, 1, 2, 2, 3, 3, 4, 4, ...)
- Gangion without Accel: indices 0 and 101, 200. All indices are duplicated except zero. (0, 101, 101, 102, 102, 103, 103, 104, 104, ...)

The PacketLossTracker works by keeping an array of all sample indices, and marching through that array.
To handle these weird cases, the PacketLossTracker class uses different sample index arrays for each board.
It's difficult to validate that the logic is correct without unit tests, so I bit the bullet and set up unit testing framework. I wanted to do this for a long time but I through it was not possible with processing. I already found some bugs thanks to these unit tests.

- Unit tests are a separate processing sketch that only runs unit tests on startup then exits.
- Located in OpenBCI_GUI_UnitTests directory
- Uses JUnit to define test classes
- Python helper script does this:
    - temporarily copy .pde files from main sketch to UnitTests sketch
    - run UnitTests sketch
    - exit(1) if tests fail
- unit tests run in CI and will fail CI if they fail

To run tests locally: `python OpenBCI_GUI_UnitTests/run-unittests.py`

image


daniellasry:
As pointed out by andrey, ganglion with accel off uses sample IDs 0, 101-200. complicated

daniellasry:
I'm opening this up for review. We discussed in the last meeting that the table was good enough for now. I cleaned it up and added the dropdown to the table. We can keep the clubhouse tak open and lower the priority.

Thanks!

daniellasry:
Also would be beneficial to merge this so we can make use of the unit testing framework

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -1373,12 +1369,12 @@ class ComPortBox {
...
 
     private LinkedList<String> getCytonComPorts() {
-        final String[] names = {"FT231X USB UART", "VCP0"};
+        final String[] names = {"FT231X USB UART", "VCP"};
         final SerialPort[] comPorts = SerialPort.getCommPorts();
         LinkedList<String> results = new LinkedList<String>();
         for (SerialPort comPort : comPorts) {
             for (String name : names) {
-                if (comPort.toString().equals(name)) {
+                if (comPort.toString().startsWith(name)) {
```
retiutut, line 1377: Not related, but should work.

retiutut:
>It's difficult to validate that the logic is correct without unit tests, so I bit the bullet and set up unit testing framework. I wanted to do this for a long time but I through it was not possible with processing. I already found some bugs thanks to these unit tests.

Awesome!

retiutut:
I've completed an initial review of the code, but I need another session to run, test, and force packet loss.

OpenBCI_GUI/ControlPanel.pde
daniellasry: There was a bug w/ Cyton. It's not always VCP0. Sometimes it's VCP1, VCP2, ...
This could be related to the issues folks were seeing when flashing many cyton boards.

daniellasry:
> I like everything except adding packetLossTreacker directly to the Board class... but it's ok

What would be a better location?

I put it on the board because the lifetime of the packetLossTracker is tied to the board.

Andrey1994:
> > I like everything except adding packetLossTreacker directly to the Board class... but it's ok
> 
> What would be a better location?
> 
> I put it on the board because the lifetime of the packetLossTracker is tied to the board.

I still think that there are or will be boards wo it, thats why adding it to the base class is not a perfect idea IMO. Maybe I would create one more class on top of Board class

"lifetime of the packetLossTracker is tied to the board."
Also probably it can be independent from board class.

But we can keep as is

retiutut:
It's freaking out using Ganglion+Wifi
Screen Shot 2020-08-16 at 9 04 19 PM

Andrey1994:
> It's freaking out using Ganglion+Wifi
> Screen Shot 2020-08-16 at 9 04 19 PM

yep, for ganglion wifi there is no magic with package id, its from 0 to 255. As I understand here ganglion ble logic applied to it instead

daniellasry:
Good to know, I cannot test ganglion wifi since mine isn't working.

Regarding:

> I still think that there are or will be boards wo it, thats why adding it to the base class is not a perfect idea IMO.

We can jsut add a function to Board like `boolean HasPacketLossTracker()`

Or we can call GetPacketLossTracker() and check for null.

Andrey1994:
"Good to know, I cannot test ganglion wifi since mine isn't working."

I think we can just copypaste it from cyton wifi

daniellasry:
@retiutut I think i fixed ganglion wifi. can you try again to confirm? Thanks

retiutut:
Screen Shot 2020-08-17 at 11 42 07 AM

daniellasry:
@retiutut one more try? Thanks!

retiutut:
WiFi range is unreal! I was able to finally get real packet loss by walking outside and putting the board behind my car (in between line of site w/ router). It works!
Screen Shot 2020-08-18 at 12 26 25 AM

daniellasry:
Hmm, that's strange. Why are you almost always losing exactly 200 packets?

retiutut:
I was really far away. I expected the Ganglion+WiFi to entirely lose connection to the local network.

daniellasry:
Gotcha. Based on that i this what is happening is that it is losing more than 200 samples and it can't accurately repost that because sample indices only go up to 200 👍

retiutut:
The console log then suggests that it lost practically every packet for a few brief seconds, likely when I put it behind the car.

Issue #805, opened 2020-06-13
Non-Smooth Connection； Ganglion ー BLED112 Dongle


Otemo:
## Problem
Every time I try to connect to the ganglion board via Bluetooth (BLED112 dongle) using the GUI, the COM port on the BLED112 dongle is updated. Windows informs me via a sound that the BLED112 Dongle COM port is refreshing.

Also, even if I stop the data streaming of the GUI and stop the session, the connection with the Ganglion board is still maintained. 
The blue diode on the Ganglion board does not flash and remains lit.

Although not smooth, we can make these connections and record data.

## Expected 
COM port doesn't refresh when connecting
Blue diode blinks when the connection is stopped

## Operating System and Version
Windows10

## GUI Version
v5.0.0-beta.0

## Running standalone app
we are running the downloaded app  and  from Processing

## Type of OpenBCI Board
Ganglion

## Are you using a WiFi Shield?
No

## Console Log

Paste any relevant text from the console window here

Otemo:
# Console Log
Console Log Started at Local Time: 2020-06-13_15-52-30
Screen Resolution: 1366 X 768
Welcome to the Processing-based OpenBCI GUI!
For more information, please visit: https://openbci.github.io/Documentation/docs/06Software/01-OpenBCISoftware/GUIDocs
TopNav: Internet Connection Not Available
Local GUI Version: 500
OpenBCI_GUI::Setup: Sample Data exists in Documents folder.
ControlP5 2.2.6 infos, comments, questions at http://www.sojamo.de/libraries/controlP5
OpenBCI_GUI::Setup: Setup is complete!
sourceList: click! 
Channel count set to 4
[DEFAULT]: BLE Devices Refreshing
ControlPanel: Found BLED112 Dongle on COM port: COM3
[DEFAULT]: BLE Devices Refreshing
ControlPanel: Found BLED112 Dongle on COM port: COM3
openbci_gui_helpers.GanglionError: Error in scan for ganglions:PORT_OPEN_ERROR
	at openbci_gui_helpers.GUIHelper.scan_for_ganglions(GUIHelper.java:69)
	at OpenBCI_GUI$BLEBox$1.run(OpenBCI_GUI.java:5502)
openbci_gui_helpers.GanglionError: Error in scan for ganglions:PORT_OPEN_ERROR
	at openbci_gui_helpers.GUIHelper.scan_for_ganglions(GUIHelper.java:69)
[...]

PR #828, opened 2020-07-02
sync gang disc threads

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -1412,7 +1412,7 @@ class ComPortBox {
 
 class BLEBox {
     private int x, y, w, h, padding; //size and position
-    private boolean bleIsRefreshing = false;
+    private volatile boolean bleIsRefreshing = false;
 
     BLEBox(int _x, int _y, int _w, int _h, int _padding) {
         x = _x;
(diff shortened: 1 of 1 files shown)


daniellasry:
1. If a user clicks the button twice, would it make more sense to halt the old thread in favor of the new one? I can image someone plugging in the dongle then pressing the button. It won't work until the previous thread is done. (and what if the old thread is stuck waiting for a timeout or something?)
1. Should we do the same thing for wifi refresh button and cyton port refresh button?

retiutut:
Does this potentially fix issue found in #805 ?

```
ControlPanel: Found BLED112 Dongle on COM port: COM3
openbci_gui_helpers.GanglionError: Error in scan for ganglions:PORT_OPEN_ERROR
	at openbci_gui_helpers.GUIHelper.scan_for_ganglions(GUIHelper.java:69)
	at OpenBCI_GUI$BLEBox$1.run(OpenBCI_GUI.java:5502)
```

Andrey1994:
1. Here it will not even create a new thread if previous is running. I think race condition between setting bleRefreshRunning in practice will not happen
2. We can and it will not hurt but not sure that its necessary

Andrey1994:
> ControlPanel: Found BLED112 Dongle on COM port: COM3
> openbci_gui_helpers.GanglionError: Error in scan for ganglions:PORT_OPEN_ERROR
> 	at openbci_gui_helpers.GUIHelper.scan_for_ganglions(GUIHelper.java:69)
> 	at OpenBCI_GUI$BLEBox$1.run(OpenBCI_GUI.java:5502)

depends on what you mean by this issue....
more likely you will not see these messages

PR #832, opened 2020-07-17
initial commit for rail percentage

Andrey1994:

feel free to improve UI)

According to my own experiments even window size == 3 is big for it. I would not recommend to set it == time series window.

### DIFF
--- a/OpenBCI_GUI/DataProcessing.pde
+++ b/OpenBCI_GUI/DataProcessing.pde
@@ -47,7 +47,7 @@ void processNewData() {
     dataProcessing.newDataToSend = true;
 
     //look to see if the latest data is railed so that we can notify the user on the GUI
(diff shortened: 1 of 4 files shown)


Andrey1994:
I would set it to smth like 1 sec because each time you change smth in electrodes connection you have to wait for window size seconds to check an effect

daniellasry:
It works, and it works consistently (if i change gain to x1, the % goes down, which makes sense)
Code looks good. `is_railed` and ` is_railed_warn` are not used anymore
The thresholds are not used either, but will probably be used for the UI later.

I think we should continue to work on the UI in this branch. I don't think it should be included in the 5.0.0 release as it is right now.

Andrey1994:
I keep is_railed and is_railed_warn because somehow they are used in other widgets(like headplot). And there are thresholds for them but its percentage which is much easier to work with...

Agreed about UI we can change it here. But keep in mind that its not for all boards

PR #840, opened 2020-07-25
Stop data stream when playback file reaches EOF

retiutut:
- [x] Simple fix
- [x] Tested

### DIFF
--- a/OpenBCI_GUI/DataSourceSDCard.pde
+++ b/OpenBCI_GUI/DataSourceSDCard.pde
@@ -135,6 +135,10 @@ class DataSourceSDCard implements DataSource, FileBoard, AccelerometerCapableBoa
 
         currentSample += numNewSamplesThisFrame;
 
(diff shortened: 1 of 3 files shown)


OpenBCI_GUI/DataSourceSDCard.pde
```diff
@@ -135,6 +135,10 @@ class DataSourceSDCard implements DataSource, FileBoard, AccelerometerCapableBoa
 
         currentSample += numNewSamplesThisFrame;
 
+        if (endOfFileReached()) {
+            stopButtonWasPressed();
+        }
+
```
daniellasry, line 141: Not good, this will repeatedly call "stopButtonWasPressed()" as long as we are at the last sample...

daniellasry: `if (endOfFileReached() || isRunning)`

is probably what you need

PR #843, opened 2020-07-27
Convert gui v4 sample data to v5 - Fixes #830

### LINKED ISSUE #830: OpenBCI GUI v5.0 freeze at "Attempt to start session" upon Sample Data opening on two different MacOs Sierra. v4.2 work.
## Problem

OpenBCI GUI v5.0 freeze at "Attempt to start session" upon Sample Data opening on MacOs Sierra

## Expected 

To start the GUI. v4.2 worked for me

macOS sierra 10.12.6

## GUI Version

5.0

## Running standalone app

Downloaded App

## Type of OpenBCI Board

No board, just playback

No

## Console Log

Paste any relevant text from the console window here

retiutut:
- [x] Fixes #830
- [x] Write some code to make sure we delete GUI v4 sample data files in Documents folder, and place new files there 
- [x] minor refactoring of related resources


OpenBCI_GUI/SessionSettings.pde
```diff
@@ -35,9 +35,9 @@
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /////////////////////////////////
-//   SoftwareSettings Class    //
+//   SessionSettings Class    //
 /////////////////////////////////
-class SoftwareSettings {
+class SessionSettings {
```
retiutut, line 40: Hey @daniellasry, as you can see here, I have renamed the vague "SoftwareSettings" tag to focus more on what it actually handles, which should focus on Session variables and widgets.

OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -223,7 +223,8 @@ static CustomOutputStream outputStream;
 public final static String stopButton_pressToStop_txt = "Stop Data Stream";
 public final static String stopButton_pressToStart_txt = "Start Data Stream";
 
-SoftwareSettings settings = new SoftwareSettings();
+GuiSettings guiSettings;
+SessionSettings settings = new SessionSettings();
```
retiutut, line 227: Keeping this variable name for `settings` in order not to introduce breaking changes across the GUI. Now that I have separated GUI-wide settings from SessionSettings, a refactor can be done soon in a separate PR.

I stopped short of doing a refactor as this was not my goal for this PR.

OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -370,44 +371,6 @@ void delayedSetup() {
     }
 }
 
-public void copyGUISampleData(){
```
retiutut, line 373: This was a global function, but now has been moved into part of instantiation of `GuiSettings` class.

daniellasry: Nice that you wanted to put the function in a class, but there's no need to make an instance of the class. It has no state. Just make the class static and make initSampleData() static. Then you can call GuiSettings.initSampleData() instead of making an instance of GuiSettings

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,72 @@
...
+                }
+            }
+            for(File file : results) {
+                Files.copy(file.toPath(),
+                    (new File(directoryName + file.getName())).toPath(),
+                    StandardCopyOption.REPLACE_EXISTING);
+            }
+        } catch (IOException e) {
+            outputError("Setup: Error trying to copy Sample Data to Documents directory.");
+        }
```
daniellasry, line 70: Ugh. Let's make this code better.

Instead of all this, use FileUtils.copyDirectory() to copy the entire sample data directory in one line. (import org.apache.commons.io.FileUtils)

https://stackoverflow.com/a/6214736

OpenBCI_GUI/OpenBCI_GUI.pde
daniellasry: You can remove the constructor from GuiSettings too.

daniellasry: Better idea: move all the contents of "GuiSettings" to DirectoryManager.pde. I created that static class for this very purpose, and its name better describes what this does.

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,72 @@
...
+
+        //Create \Documents\OpenBCI_GUI\Recordings\ if it doesn't exist
+        String recordingDirString = settings.guiDataPath + File.separator + "Recordings";
+        File recDirectory = new File(recordingDirString);
+        if (!recDirectory.exists()) {
+            if (recDirectory.mkdir()) {
+                println("OpenBCI_GUI::Setup: Created \\Documents\\OpenBCI_GUI\\Recordings\\");
+            }
+        }
```
daniellasry, line 47: Make this its own static function. It's logically separate from the rest.

daniellasry: Also, there's no need to check if recDirectory already exists. Just call recDirectory.mkdirs(). If the directories already exist, I'm pretty sure it does nothing.

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,72 @@
...
+                File[] filesFound = directory.listFiles();
+                //If this pathname does not denote a directory, then listFiles() returns null.
+                for (File file : filesFound) {
+                    file.delete();
+                }
+                println("Setup: Successfully deleted old GUI v4 sample data files!");
+            } catch (SecurityException e) {
+                outputError("Setup: Error trying to delete old GUI Sample Data in Documents folder.");
+            }
+            copySampleDataFiles(directory, directoryName);
```
daniellasry, line 29: No need to call `copySampleDataFiles` here, since it's going to get called later.

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,72 @@
...
+
+        if (guiv4_fileToCheck.exists()) {
+            //Delete old gui v4 files in Documents folder
+            try {
+                List<File> results = new ArrayList<File>();
+                File[] filesFound = directory.listFiles();
+                //If this pathname does not denote a directory, then listFiles() returns null.
+                for (File file : filesFound) {
+                    file.delete();
+                }
```
daniellasry, line 24: This can be replaced simply by:

```
for (File subFile : directory.listFiles()) {
    subFile.delete();
}
```

`results` wasn't even being used here

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,72 @@
...
+                println("Setup: Successfully deleted old GUI v4 sample data files!");
+            } catch (SecurityException e) {
+                outputError("Setup: Error trying to delete old GUI Sample Data in Documents folder.");
+            }
+            copySampleDataFiles(directory, directoryName);
+        } else {
+            //Do nothing
+        }
```
daniellasry, line 32: no need for this else clause

OpenBCI_GUI/GuiSettings.pde
retiutut: We don't use apache java library.

OpenBCI_GUI/OpenBCI_GUI.pde
retiutut: Done!

OpenBCI_GUI/DirectoryManager.pde
```diff
@@ -1,7 +1,91 @@
...
+        makeRecordingsFolder();
+    }
+
+    private static void copySampleDataFiles(File directory, String directoryName) {
+        println("OpenBCI_GUI::Setup: Copying sample data to Documents/OpenBCI_GUI/Sample_Data");
+        // Make the entire directory path including parents
+        directory.mkdirs();
+        try {
+            List<File> results = new ArrayList<File>();
+            File[] filesFound = new File(dataPath("EEG_Sample_Data")).listFiles();
```
retiutut, line 65: Build fails here because:
```
DirectoryManager.pde:65:0:65:0: Cannot make a static reference to the non-static method dataPath(String) from the type PApplet
```

daniellasry: I see. There are a few ways to fix this, but I think the simplest fix is to make DirectoryManager non-static again and create a single global instance for it in OpenBCI_GUI.pde.

Since directory manager serves more purposes now and has state data (thank you!), it's not weird to make a instance of it.

OpenBCI_GUI/GuiSettings.pde
daniellasry: ahhhh, thought we did, nevermind then

retiutut: Maybe I can do this with 1 for loop instead of 2...

retiutut:
```
try {
            File[] filesFound = new File(dataPath("EEG_Sample_Data")).listFiles();
            //If this pathname does not denote a directory, then listFiles() returns null.
            for (File file : filesFound) {
                if (file.isFile()) {
                    Files.copy(file.toPath(),
                        (new File(directoryName + file.getName())).toPath(),
                        StandardCopyOption.REPLACE_EXISTING);
                }
            }
        } catch (IOException e) {
            println("Setup: Error trying to copy Sample Data to Documents directory.");
        }
```

daniellasry: that works too

Andrey1994: I believe we can use apache.common, there is such dependency in brainflow and it should be available here too. Just import package

retiutut:
```
No library found for org.apache.commons.io
Libraries must be installed in a folder named 'libraries' inside the sketchbook folder (see the Preferences window).
OpenBCI_GUI.pde:0:0:0:0: The package “org.apache.commons.io” does not exist. You might be missing a library.
```

Issue #844, opened 2020-07-28
java.lang.UnsatisfiedLinkError: liblog.so


jacobgmartin:
## Problem

I copied all the libraries from OpenBCI_GUI (latest development version) over to the ~/sketchbook/libraries directory for Processing 4 alpha1.

Yet, I seem to have two missing libraries "liblog.so" and "processing.awt.PSurfaceAWT"

I can't seem to find any information about this error or this library on the internet.  

I'm on Gentoo.  Does anyone have any advice? Should these libraries be included in the libraries sketchbook subfolder?

## Expected 

No libraries missing.

## Operating System and Version

Linux  5.4.38-gentoo-x86_64 #2 SMP Fri May 15 04:51:57 EDT 2020 x86_64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz GenuineIntel GNU/Linux

## GUI Version

5.0.0-beta2

## Running standalone app

From processing.

## Type of OpenBCI Board

Cyton

## Are you using a WiFi Shield?

No

## Console Log

**No library found for processing.awt.PSurfaceAWT**
Console Log Started at Local Time: 2020-07-28_18-09-19
Screen Resolution: 1920 X 1080
Welcome to the Processing-based OpenBCI GUI!
For more information, please visit: https://openbci.github.io/Documentation/docs/06Software/01-OpenBCISoftware/GUIDocs
TopNav: Internet Connection Successful
Local Version: 500, Latest Version: 500
GUI is up to date!
OpenBCI_GUI::Setup: Sample Data exists in Documents folder.
ControlP5 2.2.6 infos, comments, questions at http://www.sojamo.de/libraries/controlP5
The file "/home/martin/Documents/OpenBCI_GUI/Settings/UserPlaybackHistory.json" is missing or inaccessible, make sure the URL is valid or that the file has been added to your sketch and is readable.
OpenBCI_GUI::Control Panel: Playback history file not found or other error.
null
OpenBCI_GUI::Setup: Setup is complete!
sourceList: click! 
Channel count set to 8
ControlPanel: Found Cyton Dongle on COM port: /dev/ttyUSB0
ControlPanel: Attempting to Auto-Connect to Cyton
ControlPanel: Found Cyton Dongle on COM port: /dev/ttyUSB0
Cyton AutoConnect Button: get_channel
[DEFAULT]: Attempting to open Serial/COM port: /dev/ttyUSB0
[...]

PR #845, opened 2020-07-29
fix annnoying PSurfaceAWT warning

daniellasry:


retiutut:
Nice! 

mentioned in #844

PR #850, opened 2020-08-06
add configurable gain behavior

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/ADS1299SettingsBoard.pde
+++ b/OpenBCI_GUI/ADS1299SettingsBoard.pde
@@ -244,4 +244,6 @@ interface ADS1299SettingsBoard {
     public ADS1299Settings getADS1299Settings();
     public char getChannelSelector(int channel);
     public double getGain(int channel);
+    public void setDynamicScaler(boolean val);
+    public boolean getDynamicScaler();
 };
--- a/OpenBCI_GUI/ADS1299SettingsBoard.pde
+++ b/OpenBCI_GUI/ADS1299SettingsBoard.pde
@@ -292,6 +292,6 @@ interface ADS1299SettingsBoard {
     public ADS1299Settings getADS1299Settings();
     public char getChannelSelector(int channel);
     public double getGain(int channel);
-    public void setDynamicScaler(boolean val);
-    public boolean getDynamicScaler();
+    public void setUseDynamicScaler(boolean val);
+    public boolean getUseDynamicScaler();
 };
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -238,6 +238,7 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
     protected String serialPort = "";
     protected String ipAddress = "";
     private CytonBoardMode currentBoardMode = CytonBoardMode.DEFAULT;
+    private boolean useDynamicScaler;
 
     public BoardCyton() {
         super();
@@ -247,6 +248,7 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
         // The command 'd' is automatically sent by brainflow on prepare_session
         currentADS1299Settings = new CytonDefaultSettings(this);
+        useDynamicScaler = true;
     }
 
     // implement mandatory abstract functions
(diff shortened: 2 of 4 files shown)


retiutut:
GUI @ minimum width. Looks like we have room for another button on the top right side in TopNav.
Screen Shot 2020-08-14 at 3 52 40 PM

retiutut:
Fix coming now...

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -401,7 +403,11 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
         for (int i = 0; i < exgChannels.length; i++) {
             for (int j = 0; j < data[exgChannels[i]].length; j++) {
                 // brainflow assumes a fixed gain of 24. Undo brainflow's scaling and apply new scale.
-                double scalar = brainflowGain / currentADS1299Settings.values.gain[i].getScalar();
+                double currentGain = 1.0;
+                if (useDynamicScaler) {
+                    currentGain = currentADS1299Settings.values.gain[i].getScalar();
+                }
+                double scalar = brainflowGain / currentGain;
```
daniellasry, line 410: Are we sure the scaling should happen here? Maybe it should happen in DataProcessing. Or in the timeseries widget?

I say this because if you do it here you are applying scaling to the data permanently. Should this be visual only, like the other filters? If you toggle the scaling button rapidly, you will have a mix of scaled and unscaled data displaying in timeseries.

Not to mention, you will have a mix of scaled and unscaled data in the playback file.

I think we probably want to write unscaled data to the playback file, and scaling should be a visual only button?

This means that playback data will write to file unscaled, and we will display playback data unscaled. I think this is fine. It would be complicated to have the gain scaling work in playback.

OpenBCI_GUI/TopNav.pde
```diff
@@ -126,6 +127,18 @@ class TopNav {
             smoothingButton.setHelpText("Click here to turn data smoothing on or off.");
         }
 
+        if (currentBoard instanceof ADS1299SettingsBoard) {
+            int pos_x = 0;
+            if (currentBoard instanceof SmoothingCapableBoard) {
```
daniellasry, line 132: I don't think these if statements should be nested

OpenBCI_GUI/BoardCyton.pde
Andrey1994: "Should this be visual only, like the other filters?"
I think its not like filters and its not filterring... 

"I think we probably want to write unscaled data to the playback file, and scaling should be a visual only button?"
I dont think so, also we can not store this param in playback file(now). This way playback will look and work exactly like during original run and I suppose its good

OpenBCI_GUI/TopNav.pde
Andrey1994: It affects position of the button... So I think it should be nested here

OpenBCI_GUI/ADS1299SettingsBoard.pde
```diff
@@ -292,4 +292,6 @@ interface ADS1299SettingsBoard {
     public ADS1299Settings getADS1299Settings();
     public char getChannelSelector(int channel);
     public double getGain(int channel);
+    public void setDynamicScaler(boolean val);
+    public boolean getDynamicScaler();
```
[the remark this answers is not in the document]
Andrey1994: fixed

OpenBCI_GUI/BoardCyton.pde
daniellasry: In that case the filters button area is not a good location, because the other buttons are visual only.

I think then it should be in the HardwareSettings panel. And it will force you to stop streaming before opening it, which is good.

OpenBCI_GUI/TopNav.pde
daniellasry: Ah, i think i get it now.

OpenBCI_GUI/BoardCyton.pde
Andrey1994: there is no space....

Andrey1994: I would not worry about this and playback too much....

daniellasry: Alright, I won't press the issue

PR #851, opened 2020-08-07
wip load/store settings

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/ADS1299SettingsBoard.pde
+++ b/OpenBCI_GUI/ADS1299SettingsBoard.pde
@@ -1,3 +1,6 @@
+import com.google.gson.Gson;
+import java.io.*;
+
 
 interface ADSSettingsEnum {
     public String getName();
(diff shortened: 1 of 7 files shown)


OpenBCI_GUI/ADS1299SettingsBoard.pde
```diff
@@ -155,77 +158,129 @@ enum Srb1 implements ADSSettingsEnum {
...
+    public ADS1299SettingsValues() {
+        powerDown = null;
+        gain = null;
+        inputType = null;
+        bias = null;
+        srb2 = null;
+        srb1 = null;
+        previousBias = null;
+        previousSrb2 = null;
+    }
```
daniellasry, line 181: not needed in java

OpenBCI_GUI/ADS1299SettingsBoard.pde
```diff
@@ -155,77 +158,129 @@ enum Srb1 implements ADSSettingsEnum {
...
+                fileContents.append(scanner.nextLine() + System.lineSeparator());
+            }
+            Gson gson = new Gson();
+            values = gson.fromJson(fileContents.toString(), ADS1299SettingsValues.class);
+            commitAll();
+            return true;
+        } catch (IOException e) {
+            e.printStackTrace();
+            return false;
+        }
```
daniellasry, line 255: You put the load function here but the save function is in ADS1299SettingsValues. Is there a good reason why? Could they both be in the same class?

Andrey1994:
Its readable but single line. I think there must be options in Gson to format it

OpenBCI_GUI/ADS1299SettingsBoard.pde
Andrey1994: I would put both to ADS1299SettingValues class but I would make load function static in this case . Since in processing static doesnt work I moved load function to upper class

Can move store function to the upper class too

Andrey1994:
now json file looks good and moved load/store methods to the same class

daniellasry:
Nice

OpenBCI_GUI/W_TimeSeries.pde
```diff
@@ -275,7 +302,19 @@ class W_timeSeries extends Widget {
...
             }
+            if(hardwareSettingsLoadButton.isActive && hardwareSettingsLoadButton.isMouseHere()) {
+                if (isRunning) {
+                    PopupMessage msg = new PopupMessage("Info", "Streaming needs to be stopped before loading hardware settings.");
+                } else {
+                    selectInput("Select settings file to load", "loadSettingsFileSelected");
+                }
+            }
+            if(hardwareSettingsStoreButton.isActive && hardwareSettingsStoreButton.isMouseHere()) {
+                selectInput("Save settings to file", "storeSettingsFileSelected");
```
retiutut, line 313: Should be selectOutput()

retiutut: From SoftwareSettings, you can see selectOutput() has a third argument that can be used to open a certain folder:

```

    void saveButtonPressed() {
        if (saveDialogName == null) {
            selectOutput("Save a custom settings file as JSON:",
                        "saveConfigFile",
                        dataFile(settings.getPath("User", eegDataSource, nchan)));
        } else {
            println("saveSettingsFileName = " + saveDialogName);
            saveDialogName = null;
        }
    }
```

OpenBCI_GUI/W_TimeSeries.pde
```diff
@@ -206,6 +225,8 @@ class W_timeSeries extends Widget {
 
             if(currentBoard instanceof ADS1299SettingsBoard) {
                 hardwareSettingsButton.draw();
+                hardwareSettingsLoadButton.draw();
```
retiutut, line 228: Maybe we can hide these buttons and only show if HardwareSettings are visible/active.

OpenBCI_GUI/W_TimeSeries.pde
Andrey1994: I know I forgot to change during copypaste, thanks. BTW it also works)

retiutut: It wouldn't let me save a file on Mac.

retiutut: Maybe by specifying the file path and name as an argument it opens enclosing folder automatically?

OpenBCI_GUI/W_TimeSeries.pde
retiutut: Screen Shot 2020-08-14 at 4 01 55 PM
Or we could shrink the size of each channel slightly and add them just above that, as shown in the above picture.

OpenBCI_GUI/W_TimeSeries.pde
Andrey1994: fixed selectOutput, I think we dont have to provide file, it works fine wo it...

OpenBCI_GUI/W_TimeSeries.pde
Andrey1994: I think the way it is now is fine

retiutut: Screen Shot 2020-08-14 at 4 14 39 PM
Screen Shot 2020-08-14 at 4 14 51 PM

Andrey1994: Can we just make default size(where it looks nice) a new minimum? It still will be much less than default resoultion for any display....

retiutut: This really only involves the width, let me try a increasing this value.

Andrey1994: The only problem I dont understand where this min size is configured...

retiutut: It's in the newly renamed SessionSettings, but it belongs in a new class to handle other GUI related settings. I would also like to have FPS and Expert Mode remembered after users close the GUI.

retiutut: I think the gui needs to fit in at least half of a 1080 screen.

retiutut: There's really only room for 1 more button in that top bar when I make the GUI half of a 1080 screen.

Andrey1994: added

Andrey1994: btw I even didnt know that its resizeble because there is no arrow for resize when you move mouse to the edge

retiutut: It's not going to work.

Andrey1994: "I think the gui needs to fit in at least half of a 1080 screen." Its more than 540 now... even min width currently is ~ 700. I changed it to default size which is ~half of 1920

retiutut: I'd rather move the buttons closer to the hardware settings buttons, and use longer description like in the mockup above.

retiutut: It fits!

retiutut:
image

OpenBCI_GUI/W_TimeSeries.pde
retiutut: This also resolves the issue with FFT widget seen in above image.

daniellasry:
Have you though of drawing the "load" and "save" buttons only when the hardware settings is open?

daniellasry:
Also "store" is a little unusual, "save" might be more understandable

Andrey1994:
> Have you though of drawing the "load" and "save" buttons only when the hardware settings is open?

Yeah, I think both options are fine

PR #859, opened 2020-08-19
Improve naming and organization of CI builds

daniellasry:
- Retrieve the timestamp for the commit id
- Use this timestamp to set the localGUIVersionDate
- Use the timestamp in the zip file name
- Use timestamps instead of commit IDs in build directory names
- "latest" directory will contain the most recent build. Older builds are deleted from the directory

image
image
image
image


.travis.yml
```diff
@@ -1,23 +1,25 @@
...
-    target_paths:
-      - /${TRAVIS_BRANCH}/${TRAVIS_COMMIT}
-      - /${TRAVIS_BRANCH}/latest
-    working_dir: $TRAVIS_BUILD_DIR
+
+language: python            # this works for Linux but is an error on macOS or Windows
+jobs:
+  include:	
+    - name: "Python 3.7.4 on Xenial Linux"	
+      python: 3.7.4           # this works for Linux but is ignored on macOS or Windows
```
Andrey1994, line 6: Should you add smth like os:linux to this job?

Andrey1994 (approved):

.travis.yml
daniellasry: yeah, I guess it worked anyway because linux is default?

PR #872, opened 2020-09-09
fix railed logic for gain conv

Andrey1994:


retiutut (approved):
Tested using Cyton and MarkIV headset with both gain conventions (similar reading!).

Added color indicator and tested again by disconnecting channel 1 to force railed.

Works! 🔥

PR #876, opened 2020-09-14
5.0.1 frontend improvements

retiutut:
- Remove cp5 button borders in timeseries and hsc
- lower yAxisLim font to 10 and responsively scale using average character width from font size test
- Add channel bar separator with 1 variable to adjust spacer

### DIFF
--- a/OpenBCI_GUI/W_TimeSeries.pde
+++ b/OpenBCI_GUI/W_TimeSeries.pde
@@ -100,6 +100,7 @@ class W_timeSeries extends Widget {
     private float plotBottomWell;
     private float playbackWidgetHeight;
     private int channelBarHeight;
+    public final int interChannelBarSpace = 2;
 
     private ControlP5 tscp5;
     private Button hardwareSettings;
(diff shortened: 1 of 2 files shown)


Andrey1994 (approved):

OpenBCI_GUI/W_TimeSeries.pde
```diff
@@ -776,12 +783,12 @@ class ChannelBar{
...
-        String positiveSign = limit > 0 ? "+" : "";
-        tf.setText(positiveSign+limit+"uV");
+        String s = limit > 0 ? "+" : "";
+        s += limit+"uV";
+        tf.setText(s);
         //Responsively scale button size based on number of digits
-        int n = (int)(log10(abs(limit)));
-        int padding =  n > 2 ? Math.round(map(n, 0, 10, 0, 7)) * padding_4 : 0;
-        tf.setSize(tf.autoWidth + padding, yAxisLabel_h);
+        int _width =  s.length() * 6;
```
retiutut, line 790: Here I removed the need to get the log10() by running a test on the Arial 10pt font used here. The average character width, rounded up, is 6 in this particular case.

retiutut: cp5 autoWidth was not helpful.

daniellasry (approved):

PR #888, opened 2020-09-28
Fix index.html being deleted by the eviction policy

daniellasry:


retiutut (approved):
I see that you considered alternatives and went with this. Seems like it should work. 👍

daniellasry:
Update: I decided to add index.html to github to save it.
I made it upload index.html from the build folder to s3 because that would be less fragile

Andrey1994 (approved):

PR #903, opened 2020-10-29
Fix AXR Playback mode inaccessible

retiutut:
The issue is with BatteryInfo in the Aux widget when trying to playback CSV file (Cyton, Ganglion, or AXR). Was throwing this error:

```
java.lang.ClassCastException: OpenBCI_GUI$DataSourcePlayback cannot be cast to OpenBCI_GUI$BatteryInfoCapableBoard
        at OpenBCI_GUI$W_NovaAux.<init>(OpenBCI_GUI.java:21828)
```

Here is an initial fix. Maybe there is a better way... Open to suggestions.

This PR also addresses an issue where BrainFlow throws an error when trying to connect to board, but GUI shows "success" at the bottom. Now, it does a hard `return` and exits session init process and returns UI back to control panel.

### DIFF
--- a/OpenBCI_GUI/DataSourcePlayback.pde
+++ b/OpenBCI_GUI/DataSourcePlayback.pde
@@ -360,7 +360,7 @@ class DataSourcePlayback implements DataSource, AccelerometerCapableBoard, Analo
     public Integer getBatteryChannel() {
         if (batteryChannelCache == null && underlyingBoard instanceof BatteryInfoCapableBoard) {
             try {
-                batteryChannelCache = BoardShim.get_battery_channel(BoardIds.AURAXR_BOARD.get_code());
+                batteryChannelCache = BoardShim.get_battery_channel(((BoardBrainFlow)underlyingBoard).getBoardIdInt());
             } catch (BrainFlowError e) {
                 e.printStackTrace();
             }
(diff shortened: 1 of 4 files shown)


OpenBCI_GUI/DataSourcePlayback.pde
```diff
@@ -355,6 +356,19 @@ class DataSourcePlayback implements DataSource, AccelerometerCapableBoard, Analo
         return new int[0];
     }
 
+    @Override
+    public Integer getBatteryChannel() {
+        if (batteryChannelCache == null && underlyingBoard instanceof BatteryInfoCapableBoard) {
+            try {
+                batteryChannelCache = BoardShim.get_battery_channel(BoardIds.AURAXR_BOARD.get_code());
```
Andrey1994, line 363: I think you need to use board id of underlying board instead hardcoded auraxr

OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -537,31 +537,15 @@ void initSystem() {
...
-        outputError("Failed to connect to data source fail point 1");
+        outputError("Failed to initialize board. Please check that the board is on and has power.");
+        controlPanel.open();
+        return;
     } else {
-        //initilize the GUI
+        //initilize the secondary topnav and all applicable widgets
         topNav.initSecondaryNav();
-
         wm = new WidgetManager(this);
```
Andrey1994, line 546: Why is it removed? And what/where is it?

Andrey1994 (approved):

OpenBCI_GUI/OpenBCI_GUI.pde
retiutut: This part of session init code has extra code that did nothing. It has been consolidated in place.

OpenBCI_GUI/DataSourcePlayback.pde
retiutut: I had issues accessing the method.

retiutut: Fixed. Had to cast to BoardBrainFlow

PR #905, opened 2020-11-03
Logic fix

Andrey1994:
trying to fix broken logic for start stop stream in case if exception occured.

@retiutut need to test cases like:

* trying to connect to novaxr emulator when emulator is not running
* connect to emulator(prepare_session) and disable(cntr +c) emulator and try to start streaming
* disable emulator before stop_stream
* etc

### DIFF
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -640,53 +640,52 @@ void startRunning() {
     verbosePrint("startRunning...");
     output("Data stream started.");
 
-    dataLogger.onStartStreaming();
-
     // start streaming on the chosen board
     currentBoard.startStreaming();
-    isRunning = true;
-
-    // todo: this should really be some sort of signal that listeners can register for "OnStreamStarted"
-    // close hardware settings if user starts streaming
-    w_timeSeries.closeADSSettings();
-
-    streamTimeElapsed.reset();
-    streamTimeElapsed.start();
-    sessionTimeElapsed.resume();
+    isRunning = currentBoard.isStreaming();
+    if (isRunning)
+    {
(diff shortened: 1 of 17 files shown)


OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -621,56 +621,64 @@ void initFFTObjectsAndBuffer() {
...
+    output("OpenBCI_GUI: stopRunning: stop running...");
+    currentBoard.stopStreaming();
+    isRunning = currentBoard.isStreaming();
+    if (!isRunning) {
         output("Data stream stopped.");
-
-        streamTimeElapsed.stop();
-        sessionTimeElapsed.suspend();
+        try {
+            streamTimeElapsed.stop();
```
retiutut, line 654: When haltSystem is called during session init, this causes another exception related to this timer. Fixed by only executing this code block if already running.

I might try to remove the variable `isRunning`.... I like `currentBoard.isStreaming()` much better.

retiutut (changes requested):
I will attempt to remove the `isRunning` variable , hoping it will clean up the code. Start/Stop Stream feels like spaghetti code, and it would be better if it was all handled in a button callback (a fix for another time).

OpenBCI_GUI/OpenBCI_GUI.pde
Andrey1994: I know, but its safe to keep it under try catch and doesnt hurt

Andrey1994: "Start/Stop Stream feels like spaghetti code, and it would be better if it was all handled in a button callback (a fix for another time)."
agreed I even tried it. But stopStream is called from a few unexpected places, so it was faster to keep these 2 functions

PR #919, opened 2020-11-17
rename novaxr

Andrey1994:


Andrey1994:
ci still doesnt work but it didnt work before this PR too so it doesnt matter

retiutut (approved):

PR #926, opened 2020-11-27
Enable streaming board to use Cyton Daisy source

jstuever:
This change enables the Cyton Daisy source when using Streaming board.


Andrey1994 (approved):

retiutut (approved):
Thanks! Solid contribution!

PR #931, opened 2020-12-07
5.0.2 additional UI improvements

retiutut:
- Add channel select to FFT widget
- ~~Remove Gain Convention Toggle Button https://github.com/OpenBCI/OpenBCI_GUI/pull/931/commits/7d6ac69805d41398942d73d7575b160a3a294cda~~

ADS1299 boards already use dynamic scaler by default. So, just remove the button.

- Fix Cyton Analog/Accel/Digital button behavior, set these buttons to be switches and drive button using board


OpenBCI_GUI/BoardCyton.pde
```diff
@@ -348,7 +348,7 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
     @Override
     public boolean canDeactivateAnalog() {
-        return false;
+        return true;
```
retiutut, line 351: We can deactivate analog mode on Cyton board.

retiutut: Maybe this was correct and should not be changed?

retiutut: @Andrey1994

retiutut: I changed it to true since the nearby functions are used:

```
    @Override
    public boolean isAnalogActive() {
        return getBoardMode() == CytonBoardMode.ANALOG;
    }

    @Override
    public void setAnalogActive(boolean active) {
        if(active) {
            setBoardMode(CytonBoardMode.ANALOG);
        }
    }

    @Override
    public boolean canDeactivateAnalog() {
        return true;
    }
```

retiutut: >    @Override
    public void setAnalogActive(boolean active) {
        if(active) {
            setBoardMode(CytonBoardMode.ANALOG);
        }
    }

Actually reading this again I see that it never really deactivates the mode.

retiutut: My mistake. Fixing now...

Andrey1994 (approved):

PR #947, opened 2021-01-26
Session settings fix jan2021

### LINKED ISSUE #942: Several errors in Settings through load/save/default
## Problem
@see pdf 
[bugSetting.pdf](https://github.com/OpenBCI/OpenBCI_GUI/files/5819943/bugSetting.pdf)
1- default settings when OpenBCI is started 
2- change parameters in upper options, Time series options included active channels, FFT option
3-save settings in the json file ( extension changed below from .json to .txt to be loaded in this github) 
[CytonUserSettings_000.txt](https://github.com/OpenBCI/OpenBCI_GUI/files/5819973/CytonUserSettings_000.txt)
4- restore default from setting menu:  
Upper options are not restored to default (Notch,BO filter, Smoothing (gain not tested ))
In Time series:  vertical scale is restored,BUT Window option is overriden by vertical scale value,  active channels are not restored,  
In FFT : values are restored correctly. 
5- Close then restart openGUI : same as initial 
6- Load the previously setting saved  in step 3 
Upper options are not restored to default (Notch,BO filter, Smoothing (gain not tested ))
In Time series:  vertical scale is restored,BUT Window option is overriden by vertical scale value,  active channels are not restored,  
In FFT : values are restored correctly. 

## Expected 
Must be correctly saved and restored.

Win10 Pro

## GUI Version
V5.0.1

## Running standalone app
Standalone .exe 

## Type of OpenBCI Board

Cyton

No

## Console Log
SessionSettings: loadConfigFile: User selected C:\Users\plaurent\Documents\OpenBCI_GUI\Settings\CytonUserSettings_000.json
LoadGUISettings: Applied Widget 0 to Container 4
LoadGUISettings: Applied Widget 1 to Container 3
LoadGUISettings: Applied Widget 2 to Container 9
LoadGUISettings: Expert Mode Off
Networking: Protocol mode set to Serial. Stopping network
Apply Serial Networking Mode
[SUCCESS]: Settings Loaded!

retiutut:
Fixes #942

### DIFF
--- a/OpenBCI_GUI/FilterEnums.pde
+++ b/OpenBCI_GUI/FilterEnums.pde
@@ -1,19 +1,30 @@
 public enum BandStopRanges
 {
-    Sixty(60.0d),
-    Fifty(50.0d),
-    None(null);
+    Sixty(0, 60.0d),
+    Fifty(1, 50.0d),
+    None(2, null);
 
+    private int index;
     private Double freq;
 
     private static BandStopRanges[] vals = values();
  
-    BandStopRanges(Double freq) {
+    BandStopRanges(int index, Double freq) {
+        this.index = index;
         this.freq = freq;
+    }
+
+    public int getIndex() {
+        return index;
     }
  
     public Double getFreq() {
         return freq;
+    }
+
+    public static BandStopRanges getByIndex(int i)
+    {
+        return vals[i];
     }
 
     public BandStopRanges next()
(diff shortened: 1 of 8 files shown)


OpenBCI_GUI/FilterEnums.pde
```diff
@@ -1,21 +1,32 @@
 public enum BandStopRanges
 {
-    Sixty(60.0d),
-    Fifty(50.0d),
-    None(null);
+    Sixty(0, 60.0d),
```
Andrey1994, line 3: why is it changed and why 0?

Andrey1994 (approved):

OpenBCI_GUI/FilterEnums.pde
retiutut: I added an “index” to these enums so I wouldn’t have to use ‘ordinal’. I believe this follows Java style guide.

PR #952, opened 2021-02-08
update brainflow jar

Andrey1994:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v5.0.4
+
+### Improvements
+* Update BrainFlow library to version that includes a marker channel
+
 # v5.0.3
 
 ### Improvements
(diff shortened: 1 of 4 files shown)


retiutut:
@Andrey1994 I think the BrainFlow library was deleted instead of updated.

Andrey1994:
https://github.com/OpenBCI/OpenBCI_GUI/blob/brianflow_new_ver/OpenBCI_GUI/libraries/brainflow/library/brainflow.jar here is this file

retiutut:
@Andrey1994 I can see on Github Desktop that the file was modified, but for some reason Github in browser shows this:

Screen Shot 2021-02-08 at 1 35 08 PM

It's fine and the file was indeed updated.

_______

@Andrey1994 Do you think any other channels would be added to BrainFlow in the next year or two? What happens to this code/PR if another channel is added later?

Andrey1994:
I dont think that smth else will be added to all boards

retiutut:
@Andrey1994 Last questions:

- To confirm, this code and BrainFlow update avoids a breaking change?
- We are clear now to utilize this new BrainFlow feature in a GUI widget/feature? Maybe after restoring Focus widget?

PR #956, opened 2021-03-14
Gui memory fix

retiutut:
This might make some parts of the GUI look wrong. This needs to be fixed before the next major release.


retiutut (commented):
Looks good!

PR #957, opened 2021-03-15
Hw settings march2021

### LINKED ISSUE #954: GUI v5.0.3 fails to send/update/maintain hardware settings
## Problem

I'm trying to record EEG and EOG simultaneously following instructions from [this official tutorial](https://github.com/openbci-archive/Docs/blob/master/Tutorials/01-Cyton_Getting%20Started_Guide.md#4-launch-the-gui-and-adjust-your-channel-settings). I use Cyton+Daisy but in this case I just switch to 8 channels recording. The signal is then sent through LSL and intercepted by [LabRecorder](https://github.com/labstreaminglayer/App-LabRecorder), since I need to synchronize it with a stream of markers.
My configuration:
A1 - bottom SRB
A2 - bottom BIAS
EOG - N1P (bipolar: top pin + bottom pin)
EEG - N8P

Because I dont't need the rest of the channels and I don't want my EOG channel to refer either to SRB or Bias, I go to Hardware settings and first disable unused channels 2-7, and then for channel 1 I switch off both SRB2 and Bias. I hit "Send" and get a message that it was sent to my board. I go back to Time series and it seems to work fine. However, after ~2-3 minutes the disabled channels (usually 3 and 4) start to send some high frequency noise. After another ~60 seconds all channels (including EEG and EOG) produce only HF noise. Updating settings at this stage is impossible. I tested this with GUI v4.2 and everything worked just fine. GUI 5.0.3 works without problems when I don't touch Hardware settings (but I really need to...).

## Expected 

Hardware settings are sent to Cyton and maintain until the end of current session or until they are updated (as it is in v4.2).

Windows 10

## GUI Version

5.0.3

## Running standalone app

I'm running the downloaded app

## Type of OpenBCI Board

Cyton/Cyton+Daisy

No

retiutut:
Fixes #954

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ libBoardController.so
 libDataHandler.so
 libGanglionLib.so
 libGanglionScan.so
+libunicorn.so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Handle paths with spaces on Linux Standalone GUI #916
 * Allow Expert Ganglion Users to send square wave commands via keyboard #950
 * Show Send Custom Hardware Command UI for Cyton Expert Mode in Hardware Settings
+* Improve Hardware Setting UX/UI for ADS1299 boards #954
 
 ### Bug Fixes
 * Clean up GUI code to fix Processing/JVM memory issue causing crash #955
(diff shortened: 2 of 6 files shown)


retiutut (commented):
Looks good! Self code review and tested UX/UI for bugs. Needs more testing in beta to confirm.

PR #967, opened 2021-04-30
fix initial settings done

pirelaurent:
This is to fix a bug in settings . 
See file pirla_README.MD   
( first time i do a contribution, hope no mistake )


retiutut:
I'm concerned that this pull request does not address any documented issues. It may not be merged.

retiutut:
I recommend making an issue that states what the problem is. We have discussed completely refactoring software settings to make the code more simple.

pirelaurent:
Yes, settings is much more clear but the bug previously described by #942
is not corrected.
I reopen a #967 on same pb.
I hope you will read the analysis and solution.
My pull request was designed for that.
Apologize not to have reopen a bug before.
Please, have a look .

Le lun. 3 mai 2021 à 18:49, Richard Waltman ***@***.***> a
écrit :

Issue #969, opened 2021-05-03
Several options ( Layout, Notch, Bandpass Filter, widgets,etc. ) in DefaultSettings.json are not applied by the app.


pirelaurent:
## Problem
DefaultSettings file is overwritten by app at every startup.  Whatever you have saved as the default config, it is lost. 
A main drawback motivation is for example Notch. If you leave in a country at 50hz, you must change by hand the notch each time even if you saved previously on defaultsettings (or you patch the file ) .  
 
## Expected 
App must take care of the current defaultSettings.json values at startup.
I have proposed a pull request #967 with all explanations ( pirela_README.md)  and solution (quite simple since settings refactoring :) )  
Was closed probably because no relation with a bug.  Now, there is one. 
[pirla_README.md](https://github.com/OpenBCI/OpenBCI_GUI/files/6416685/pirla_README.md)

## Operating System and Version

WIndows

## GUI Version

5.0.4 

## Running standalone app

visual studio code or standalone

## Type of OpenBCI Board

Cyton

## Are you using a WiFi Shield?

No

## Console Log

Paste any relevant text from the console window here

retiutut:
>App must take care of the current defaultSettings.json values at startup.
I have proposed a pull request #967 with all explanations ( pirela_README.md) and solution (quite simple since settings refactoring :) )
Was closed probably because no relation with a bug. Now, there is one.
pirla_README.md

I will review this carefully, and I have assigned this issue to myself. Thank you for bringing this up!

pirelaurent:
Hello Richard,
 thanks to take this in account.
 As i'm new in the contribution process with fork and pull request, i'm not
sure to have pull the right last one (may be forget a prior commit of my
branch fixSettings)
But, in any case, the pirla_README.md has all the (tiny) code to apply.
I had leaved a question inside about galea and streaming code that are
excluded of initial settings. But probably, we must load stored settings
for them too.
Ready to exchange if this help.

have a nice day
Pirla , from France






Le mar. 4 mai 2021 à 18:36, Richard Waltman ***@***.***> a
écrit :

retiutut:
@pirelaurent I will see if this can be an easy fix, but I proposed a much broader and general fix to these kinds of issues, found here:

#970

pirelaurent:
Hello,
I saw #970 to generalize saving of widgets configuration, but my  #969 is
not on the same level, it's on the dynamic of settings, not on the content.
[...]

Issue #964, opened 2021-03-31
Clarify GUI system requirements


retiutut:
## Problem

GUI requires Windows 8 or 10. Console logs only show "Windows". It should be possible to print the windows version on app start.

## Expected 

Warn users who are running Windows 7 with a popup window, since users will not be able to connect to board. Requires using a different computer or upgrading to Windows 8.1 or 10.

Also, could help to share system requirements on openbci.com/downloads

## Operating System and Version

Windows 7, 8, and 10

## GUI Version

GUI 5.0.4

## Running standalone app

Standalone

## Type of OpenBCI Board

All boards

## Are you using a WiFi Shield?

No

## Console Log

```
Console Log Started at Local Time: 2021-03-31_12-38-43
Screen Resolution: 1920 X 1080
Operating System: Mac <----- Display detailed version of Windows at this point on GUI start.
Welcome to the Processing-based OpenBCI GUI!
```

retiutut:
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SystemUtils.html

retiutut:
https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/SystemUtils.html

Issue #924, opened 2020-11-21
Restore Focus Widget in the GUI


retiutut:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Restore the Focus widget using information from BrainFlow metric. Front end can remain the same for now. Important thing is to have a Focus widget with modest accuracy.

Restore Focus Widget output to the Networking widget for all networking modes.

**Describe alternatives you've considered**
This was removed as per a team decision related to the accuracy of the old Focus widget.

**Additional context**
The backend work for this issue has been completed! BrainFlow now has a metric that uses a simple classifier. 🛠✅

Need to update Docs also once this is coded and tested in the GUI. Specifically, the GUI Widget Guide.

retiutut:
- [ ] Make new widget
- [ ] Include channel select bar in the widget
- [ ] Add output to networking widget for all protocols that outputs the same as the old widget
- [ ] Update GUI Widget Guide Doc or any other references to the Focus widget in the GUI

retiutut:
>Add output to networking widget for all protocols that outputs the same as the old widget

No guarantee here. Going to do what seems simplest.

retiutut:
Keeping the output of the Focus Widget Datatype to 1's and 0's, to make this easier for children and students to learn about binary, rather than expecting ALL users to code a threshold on the other end of the networking stream. Threshold can be set via the new Focus Widget.

retiutut:
![Screen Shot 2021-05-28 at 7 09 34 PM](https://user-images.githubusercontent.com/25914123/120051899-3eb06e80-bfe8-11eb-8ff6-fe69f4dac13c.png)

retiutut:
Last task is to update the official OpenBCI Documentation.

I should also update the Instructable I made: https://www.instructables.com/Send-Focus-Data-From-OpenBCI-GUI-to-Arduino/

I'm happy to keep the same Arduino code to make the above work.

PR #975, opened 2021-05-29
GUI 5.0.5

retiutut:
# v5.0.5

### Improvements
* Implement Focus Widget using BrainFlow Metrics! #924
* Throw a popup if users are are running an old version of Windows operating system. GUI v5 supports 64-bit Windows 8, 8.1, and 10. #964
* Throw a popup if Windows users are using 32-bit Java and Processing. #964
* Set Networking Widget default baud rate for Serial output to 57600

### Bug Fixes
* Fix Y axis Autoscale in TimeSeries when all values are less than zero. Example: Cyton with filters off
* Gracefully handle cases when Cyton or Cyton+Daisy users want to use 8 or 16 channels #954
* Update Save Session Settings success message. Session settings are no longer auto-loaded on Session start. #969
* Session settings are no longer auto-saved when system is halted #969


retiutut (commented):
Self-reviewed twice. Approved.

PR #994, opened 2021-09-02
5.0.7 Networking HotFix

retiutut:
- Fix streaming of multiple data types


retiutut (commented):
Tested as best as I could using Synthetic and Cyton data modes in the GUI. Sampling rates were checked using the Networking test kit in this repo.

Initial code review passing. Removes silly code, and appears to be functional.

retiutut (commented):
Second self code review passes. Hoping for the best with this code, since it falls short of doing a complete refactor of the W_Networking.pde file.

Issue #997, opened 2021-09-13
GUI-wide Settings class


retiutut:
**Is your feature request related to a problem? Please describe.**
Not related to a problem per-say.

**Describe the solution you'd like**
A handful of features should persist once the GUI has been run for the first time and then in each subsequent app instance. 

**Example: Expert Mode.*

Making this class and doing the two above settings would be a great start.

**Data can be stored along with other settings in USER/Documents/OpenBCI_GUI/Settings**

**Describe alternatives you've considered**
Session settings should pertain only to the session.

**Additional context**
Need to make this class follow the pattern set forth in the the hardware settings class. DO NOT use the pattern set forth in the Session Settings class, since this needs to be refactored.

retiutut:
Relate to #970 in that there may be some residual settings that need to be in this new GUI-wide class.

retiutut:
Cyton Data smoothing is a session-specific setting and should remain in sessionSettings.pde class.

retiutut:
Testing is passing!

PR #998, opened 2021-09-17
GUI-wide settings class

retiutut:
Add GUI-wide settings class to keep certain settings across sessions and app starts #997

Uses pattern established by ADS1299SettingsController.pde which uses GSON and a specific class for storing data values.

### DIFF
--- a/OpenBCI_GUI/GuiSettings.pde
+++ b/OpenBCI_GUI/GuiSettings.pde
@@ -1,0 +1,116 @@
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.io.*;
+
+
+interface GuiSettingsEnum {
+    public String getName();
+}
+
+enum ExpertModeEnum implements GuiSettingsEnum {
+    ON("Active", true),
+    OFF("Inactive", false);
+
+    private String name;
+    private boolean val;
+
+    ExpertModeEnum(String _name, boolean _val) {
+        this.name = _name;
+        this.val = _val;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    public boolean getBooleanValue() {
+        return val;
+    }
+}
+
+public class GuiSettingsValues {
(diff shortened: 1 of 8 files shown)


OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,116 @@
...
+            return true;
+        } catch (IOException e) {
+            e.printStackTrace();
+            outputWarn("OpenBCI_GUI::Settings: Error saving GUI-wide settings to file. Please make an issue on GitHub.");
+            return false;
+        }
+    }
+
+    //Call this method at the end of GUI main Setup in OpenBCI_GUI.pde to make sure everything exists
+    public void applySettingsToFrontEnd() {
```
Andrey1994, line 102: it seems like such methods should be in another class(in classes for particular setting if you already created them), maybe the interface for settings can have a method like `applySettingsToFrontEnd` and these methods should be called from the *main* applySettings method

Andrey1994 (approved):

OpenBCI_GUI/GuiSettings.pde
retiutut: Ok giving this a try. Pushing a new commit to address this.

retiutut: I'm getting lots of null pointers trying to move these methods and still maintain functionality.

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -0,0 +1,116 @@
...
+        topNav.configSelector.toggleExpertModeFrontEnd(values.expertMode.getBooleanValue());
+    }
+
+    public void setExpertMode(ExpertModeEnum val) {
+        values.expertMode = val;
+        saveToFile();
```
retiutut, line 110: Changing a setting will also need to update the settings file. Putting this method in the GuiSettingsValues didn't seem to work.

Issue #1001, opened 2021-09-28
GUI shows as a plain white screen (doesn't load) on Mac


pietrog30:
## Problem

I just purchased the Ultracortex Mark IV Setup. I just downloaded the GUI on my Mac, but after launching it, it just shows a white page. I tried removing and reinstalling the application like 4 times and restarting the Mac as well. Nothing seems to help, and I am unable to view the brain data.

<img width="1030" alt="Screen Shot 2021-09-28 at 1 32 47 PM" src="https://user-images.githubusercontent.com/26072858/135167565-5e5a7cec-c430-450b-9cdf-e8b9aa3221c0.png">

## Expected 

The GUI should load as it does in any tutorial I've seen.

## Operating System and Version

macOS Big Sur - Version 11.6

## GUI Version

Version 5 (5.0.6)

## Running standalone app

Running the downloaded app

## Type of OpenBCI Board

Cyton

## Are you using a WiFi Shield?

No

retiutut:
Deploying fix that has passed testing tonight!

Issue #992, opened 2021-08-23
Show Popup when new GUI version is available


retiutut:
**Is your feature request related to a problem? Please describe.**
Add a popup for all users when a newer version of the GUI is available.

**Describe the solution you'd like**
Hopefully this wouldn't be too annoying? Seems like it could help users stay up to date and increase the number of users who are testing the latest version of the GUI.

**Describe alternatives you've considered**
There is an update button already. User need to hover over this, click it, or check console log.

**Additional context**
 Plenty of other applications use popups to alert to new versions in 2021.

retiutut:
![Screen Shot 2021-09-03 at 5 25 26 PM](https://user-images.githubusercontent.com/25914123/132071585-2b49c808-14ee-4ee5-b633-606567f8a4e6.png)

Easier and less annoying fix for now is to just display a `outputWarn()` in the footer.

Issue #990, opened 2021-08-19
OpenBCI GUI v5.06 opens blank screen - macOS Catalina


Daisykh:
## Problem

When updated from 5.05 to 5.06, the OpenBCI GUI fails to open and displays a blank grey screen. v5.05 opens as expected with full functionality to connect to a Cyton board. I have tried 3 times to update and each time have had the same issue with 5.06 causing me to uninstall and reinstall 5.05. Image of GUI with version number below:
![image](https://user-images.githubusercontent.com/29408184/130103196-584cd5be-13dd-4946-ad39-eb53e4bb4034.png)

## Expected 

Opening the app should display the GUI as it does with 5.05. Image of previous version working below:
![image](https://user-images.githubusercontent.com/29408184/130103722-df55661c-633d-465d-acc6-622d383c565a.png)

## Operating System and Version

macOS Catalina, v10.15.7
Macbook Pro 2019

## GUI Version

5.06
## Running standalone app

Downloaded GUI app
## Type of OpenBCI Board

Cyton board

## Are you using a WiFi Shield?

No

## Console Log

n/a, console log doesn't appear

retiutut:
https://github.com/OpenBCI/OpenBCI_GUI/issues/987

This is likely a duplicate issue that will have a fix in the next version of the GUI. To confirm, you could follow the same steps as in the other thread.

Daisykh:
Ahh! Yes, looks like the same issue even though I'm on an intel macbook pro and not M1! Here's the results:

1. Running via processing3, no changes made and results in the blank screen. Here's the console output:
![image](https://user-images.githubusercontent.com/29408184/130108867-f139fa81-ea1c-4152-9a48-0a7501804c39.png)

2. Commented out `pixelDensity` and it now opens! Here's the console log output that includes the screen size and High-DPI detection of my macbook:
![image](https://user-images.githubusercontent.com/29408184/130109230-3a059db8-1a08-4389-9ca4-2e7060d323b5.png)

3. I also tried moving `pixelDensity(displayDensity())` into the `settings` function and it produced the same results as commenting it out entirely. Here's a screen grab:
![image](https://user-images.githubusercontent.com/29408184/130110201-f9e8974e-49b7-433b-8030-fb2c2e364730.png)

This issue can be closed as a duplicate and I can relay my findings in #987 if it would be helpful.

retiutut:
[...]

Issue #987, opened 2021-08-12
OpenBCI GUI v5.0.6 shows only a blank window upon startup on my MacBook M1


ScottThomasMiller:
## Problem

When I open the GUI, it displays only a blank window.  

![image](https://user-images.githubusercontent.com/73297684/129265832-47d7160b-1ab6-4738-8ff4-98e3149040cf.png)

## Expected 

I expect to see the usual startup window.

![image](https://user-images.githubusercontent.com/73297684/129265162-d22fad27-2f46-4cb8-81d5-47d472c217f9.png)

## Operating System and Version

macOS Big Sur 11.5.1 (20G80) on MacBook M1

## GUI Version

5.0.5

## Running standalone app

downloaded app 

## Type of OpenBCI Board

Cyton+Daisy

## Are you using a WiFi Shield?

No

## Console Log

N/A

ScottThomasMiller:
Prior versions work fine.  I dropped back to v5.0.4, which seems to run without any issues.

ScottThomasMiller:
Sorry about the confusion.  The non-working version is 5.0.6.  The working version is 5.0.5.

retiutut:
@ScottThomasMiller Thank you for making this issue. 

Can you try and locate the Console Log file for 5.0.6 in `/USER/Documents/OpenBCI_GUI/Console_Data` and share it here?

If the window opens, I have a feeling there might also be a Console Log and specific info on what could be causing the issue.

ScottThomasMiller:
It does not produce a console log.  I am looking in /Users/scottmiller/Documents/OpenBCI_GUI/Console_Data.

retiutut:
I was able to run 5.0.6 on a 2014 Mac Mini running Big Sur 11.5.2. Tagging this issue as hardware related.

Also tagging an upstream issue attempting to be tackled by the Processing team in Processing 4. https://github.com/processing/processing4/issues/128

retiutut:
@ScottThomasMiller Would you be willing to help debug this further by running the GUI from Processing?

It's interesting that we are running Processing 3, and 5.0.5 **does work**.....

Maybe it broke by implementing the following fix https://github.com/OpenBCI/OpenBCI_GUI/issues/968 which sets pixel density...

If you are available to help I can provide minimal instructions on how to test this.

ScottThomasMiller:
[...]

PR #1002, opened 2021-09-28
Development 5.0.7

retiutut:
# v5.0.7

### Improvements
* Show info in footer when a new version of the GUI is available #992
* Add GUI-wide settings class to keep certain settings across sessions and app starts #997
* Remove 30 second window option from Focus widget

### Bug Fixes
* Fix GUI not running on some Macs due to high-dpi screen code #987 #990 #1001
* Fix streaming multiple data types over LSL #971


retiutut (commented):
Passing self code review. Modified GUI update button logic to account for practically all scenarios. Decided to store a single Boolean as a global variable so users can check for Update at a later time using the button.

PR #1015, opened 2021-12-07
Add Brainflow Streamer to GUI

### LINKED ISSUE #1007: Add BrainFlow Streamer Output to GUI
Users would like to make recording files that can also be used directly with BrainFlow API.

Add the "BrainFlow  Streamer" feature to the GUI which would allow users to output all data from all OpenBCI boards to either a separate application or a file.

This could be useful for some users who would like to use the GUI to check signals and view live data before/during making a recording file that could be used again during offline processing using any BrainFlow binding.

Backend for this feature is relatively easy. Frontend fix involves adding UI to the control panel for all boards.

GUI CSV files and BDF files can still be saved at the same time!

retiutut:
Fixes #1007


OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -735,10 +737,10 @@ void initFFTObjectsAndBuffer() {
 
 void startRunning() {
     // start streaming on the chosen board
+    dataLogger.onStartStreaming();
```
retiutut, line 740: This was very carefully moved. In order for BrainFlow files to be saved next to OpenBCI files, the output directory needed to be set first before startStream() for all boards.

Andrey1994 (approved):

PR #1032, opened 2022-03-06
fix button status in networking widget

Andrey1994:


retiutut:
Seems like it should fix the error! Thanks @Andrey1994

retiutut (approved):

PR #1033, opened 2022-03-07
Set Cyton Data Smoothing to True by Default and Change Button Text

### LINKED ISSUE #1026: Apply smoothing by default
## Problem

Inconsistent waveform update in GUI (known issue) when COM driver latency not changed from 16ms to 1ms and smoothing option is not on.

## Expected 

The software smoothing option should be on by default since that mostly fixes the problem. A "pro-tips" pop-up the first time the user starts the software can alert them to the fact they can shave a few hundred ms off the total latency (in addition to fixing the smoothness problem) by changing the latency setting.

Win10

## GUI Version

5.09

## Running standalone app

downloaded binary

## Type of OpenBCI Board

Cyton

No

### LINKED ISSUE #1027: Smoothing button text is unclear
## Problem

It is not clear if the button that says "Smoothing On" is reporting the current state of the button, or telling me what will happen if the button is pressed.

## Expected 

If the text said "Turn Smoothing On", and the background changed from blue to green, that would give a clear indication of its state.

Win10

## GUI Version

5.09

## Running standalone app

binary

## Type of OpenBCI Board

Cyton

No

retiutut:
Fixes #1026 
Fixes #1027 

Update CI/CD to Processing 4 Beta 7 for all OS.


retiutut:
@Andrey1994  It looks like app signing may be failing for both Windows and Mac builds. I know we have looked at this before, might be time to look at it again and maybe start from scratch.

OpenBCI_GUI/GuiSettings.pde
```diff
@@ -30,45 +31,57 @@ enum ExpertModeEnum implements GuiSettingsEnum {
...
+                tempNumLines++;
             }
             Gson gson = new Gson();
-            values = gson.fromJson(fileContents.toString(), GuiSettingsValues.class);
+
+            if (linesByDefault == tempNumLines) {
+                values = gson.fromJson(fileContents.toString(), GuiSettingsValues.class);
+                println("OpenBCI_GUI::Settings: Found and loaded existing GUI-wide Settings from file.");
+            } else {
+                println("OpenBCI_GUI::Settings: Incompatible GUI-wide Settings found. Creating new file and resetting defaults.");
```
Andrey1994, line 82: I dont get the idea of validating json using number of lines, its not good.
As an alternative you can check that json is valid by loading it and checking required keys

retiutut: @Andrey1994 

```
    private boolean validateJsonKeys(String stringToSearch) {
        List<String> foundKeys = new ArrayList<String>();
        Gson gson = new Gson();
        Map<String, Object> map = gson.fromJson(stringToSearch, new TypeToken<Map<String, Object>>() {}.getType());
        map.forEach((x, y) -> foundKeys.add(x));

        Collections.sort(valueKeys);
        Collections.sort(foundKeys);

        boolean isEqual = valueKeys.equals(foundKeys);

        return isEqual;
    }
    ```

retiutut: It's pretty nice to be able to start using modern Java syntax, Collections, and iterators.

Andrey1994 (approved):

PR #1034, opened 2022-03-11
Update Band Power Widget and Add Average Band Power data type to Networking

retiutut:
- Update Band Power Widget
- Add Average Band Power data type to Networking
- Set Default networking protocol to UDP (since this can be more helpful now in 2022, with the advent of IoT).
- Fix backgrounds behind dropdowns in Networking Widget
- De-prioritize TimeSeries data type by moving it farther down the list in Networking widget data type dropdowns
- Fix error in Headplot widget found using Processing 4 where the image was null. Instead wait for doHardCalcs after screen is resized.
- Bump version to 5.1.0-alpha.5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 * Stop data stream when no data received after 5 seconds #1011
 * Revisit Ganglion Impedance widget so it behaves like new Cyton Impedance Widget #1021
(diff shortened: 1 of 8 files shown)


retiutut (commented):
Passes self code review and CI/CD tests.

Andrey1994 (approved):

PR #1038, opened 2022-04-08
Custom filters 2022

### LINKED ISSUE #988: Custom Filters per Channel
A few users, including some OpenBCI team members, have requested the feature of custom filters on a per channel basis.

https://openbci.com/forum/index.php?p=/discussion/3087/how-to-setup-custom-filters-and-time-series-graphs#latest

This is hard to implement from a UI/UX perspective as in where exactly this is supposed to fit in the GUI or TimeSeries widget.
The TimeSeries widget is already jam-packed with UI objects.

The backend code for such a feature is very easy using BrainFlow filters.

Allow users to select different filters per channel. Ex: Ch 1 5-50Hz filter, Ch 2 3-13Hz filter, Ch 3 complete custom filter

Currently the GUI uses two types of filters. Possibly allow users to type in their own frequencies for this.

This has been avoided because there isn't much room in the TimeSeries widget for any more complex UI objects.

## **Additional context**
This might be flagged as a won't fix because there are only a few users asking for such advanced functionality.

retiutut:
Fixes #988


Andrey1994 (approved):

retiutut:
Unfortunately, we will have to downgrade to Processing 3 as a result of sustained issues with Processing 4 and popup windows.

PR #1039, opened 2022-04-11
bump brainflow version

Andrey1994:

should set latency to 1ms automatically for cyton
@retiutut need to test before merging


retiutut (approved):
It works! Making a simple edit to the FTDI buffer fix popup to explain this is taken care of.

retiutut:
Actually, just going to hide the FTDI popup on Windows.

PR #1041, opened 2022-04-15
Downgrade to Processing 3

retiutut:
There were issues with the new FilterUI popup window that aren't being addressed in Processing 4. Namely, the `surface.setSize()` method would crash the entire window. Also were seeing lots of accessibilityHitTest errors that stem from Java directly. Let's downgrade to Processing 3.

Agreed plan of action while pair-coding with @Andrey1994 .


retiutut (commented):
Passing self code review.

PR #1043, opened 2022-04-20
Change Windows Priveleges and Check Privileges

retiutut:
* Update priveleges for Windows users and check if GUI has been run as Administrator

### DIFF
--- a/OpenBCI_GUI/Extras.pde
+++ b/OpenBCI_GUI/Extras.pde
@@ -47,13 +47,20 @@ private void checkIs64BitJava() {
         PopupMessage msg = new PopupMessage("32-bit Java Detected", "OpenBCI GUI v5 and BrainFlow are made for 64-bit Java (Windows, Linux, and Mac). Please update your OS, computer, Processing IDE, or revert to GUI v4 or earlier.");
     }
 }
-
 /**
-     * Determines if elevated rights are required to install/uninstall the application.
-     *
-     * @param path the installation path, or <tt>null</tt> if the installation path is unknown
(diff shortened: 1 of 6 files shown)


retiutut:
@Andrey1994 I created a new account on a Windows 10 computer that was not an Administrator, then ran the GUI from this build. It didn't seem to trigger:

```
    if (!AdministratorChecker.IS_RUNNING_AS_ADMINISTRATOR) {
        println("OpenBCI_GUI: This application is not being run with Administrator access. This could limit the ability to connect to devices or read/write files.");
    }
```

OpenBCI_GUI/Extras.pde
```diff
@@ -42,6 +48,47 @@ private void checkIs64BitJava() {
     }
 }
 
+//https://stackoverflow.com/a/23538961
+public static class AdministratorChecker
```
retiutut, line 52: Here is supposedly a method that works on more than just Windows, though it didn't seem to be triggered when running the app as a non-administrator and regular read/write access on Windows 10.

Andrey1994: I've found smth that looks reasonable: https://github.com/izpack/izpack/blob/82182fbafec83d1c5fd411dfefe57f9b09ad59c4/izpack-util/src/main/java/com/izforge/izpack/util/PrivilegedRunner.java 

lets copypaste method `public boolean isElevationNeeded(String path)` and `public boolean isAdminUser()` and print notification if any of these two returned `false`. For now only windows is needed

Andrey1994: actually these methods looks like they may work on unix too

PR #1052, opened 2022-05-03
Unified triangles in dropdowns - UX Review

retiutut:
Unify the shape and direction of triangles in dropdowns and custom buttons in:

- Channel Select in widgets
- Hide/Show channels in Filter UI
- ScrollableList Dropdowns (no change here)


retiutut (commented):
Passing self code review and review for UX team members.

PR #1053, opened 2022-05-08
Package loss notify

Andrey1994:


retiutut (approved):
Works as expected! Tested by walking varying distances throughout the house. After 5 seconds of no packets, the popup throws also that there was no new data.

Now the GUI has more feedback into poor connection:
1. Good signal
2. Some dropped packets (print to console log every time)
3. Many dropped packets (now with an alert at the bottom of the GUI from this PR)
4. Lost connection for X seconds (throw popup for no new data)

PR #1055, opened 2022-05-11
change default railed percentage

Andrey1994:

### DIFF
--- a/OpenBCI_GUI/Extras.pde
+++ b/OpenBCI_GUI/Extras.pde
@@ -413,10 +413,17 @@ class DataStatus {
             int endPos = data.length;
             int startPos = Math.max(0, endPos - nPoints);
 
+            boolean is_straight_line = true;
+            if (!currentBoard.isStreaming()) {
+                is_straight_line = false;
+            }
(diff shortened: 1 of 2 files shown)


OpenBCI_GUI/Extras.pde
```diff
@@ -413,11 +413,18 @@ class DataStatus {
...
+            if (!currentBoard.isStreaming()) {
+                is_straight_line = false;
+            }
             float max = Math.abs(data[startPos]);
-            for (int i = startPos; i < endPos; i++) {
+            for (int i = startPos + 1; i < endPos; i++) {
                 if (Math.abs(data[i]) > max) {
                     max = Math.abs(data[i]);
                 }
+                if ((Math.abs(data[i - 1] - data[i]) > 0.00001) && (Math.abs(data[i]) > 0.00001)) {
```
retiutut, line 425: @Andrey1994 So raw data, even when subtracted against any other point of raw data, is always greater than 0?

retiutut (approved):
Tested with Cyton with no electrodes attached and then touching pins with fingers, removing fingers, then touching again, etc.

Works!

Issue #705, opened 2020-03-27
Feature Request: GUI dark mode


projectmentor:
Is it possible to have an alternate css color scheme? Something like solarized dark for people with TBI and concussion it is difficult to visually process bright white background.  I just started here and would be willing to help once i get  up to speed on the documentations.

retiutut:
@projectmentor I am highly in favor since you have specifically requested an accommodation! Will add this to the next Software Team meeting. 💯

retiutut:
Currently, we have an alternate color scheme.

>To change to the alternate color scheme, use '{'.
https://docs.openbci.com/docs/06Software/01-OpenBCISoftware/GUIDocs#expert-mode

So, we will either adding a third theme (dark mode), or replacing the alternate color scheme with dark mode.

Thank you for requesting this!

Dorson:
Have a look at this simple example. Unlimited color themes possible:

https://dorson.github.io/CSS-Dark-Mode-and-color-switch/

MaxJackson:
I'm thinking about picking this up and working on it! Would it make sense to have a 'Dark Mode' button available in the settings drop-down menu? If so, would it also make sense to move the 'alternate color scheme' configuration option over there too?

retiutut:
Sure, but behind the scenes I think we should make a "Style" class, where the mode can change and everywhere else says `style.getWidgetBackgroundColor()` etc.

retiutut:
@MaxJackson +1 priority bump. Thank you for supporting this feature request!

MaxJackson:
@retiutut I'm debating between 

- putting the toggle functionality in the ControlPanel and having it be a setting that the user can only toggle when initializing a session
- putting the toggle functionality in with the configOptions buttons and letting the user toggle it at any time

From what I can tell the first option will be much easier to implement, since the second option will require implementing a updateNavButtonsBasedOnColorScheme() analogue touching everything with a color we want to change.

Do you have any input/considerations before I go ahead and implement the first option?

MaxJackson:
@retiutut I added a Dark Mode / Light Mode toggle to the control panel and added dark mode functionality to every widget available in the synthetic data mode: check out https://github.com/MaxJackson/OpenBCI_GUI/tree/dark-mode-toggle to see where I'm going with this.

retiutut:
It's kind of all over the place. This is why I mentioned making a style class.

MaxJackson:
[...]

PR #1063, opened 2022-05-27
Prepare foundation to add new themes

Andrey1994:


retiutut:
Use code from this PR to complete #705

Issue #1080, opened 2022-09-02
Issue when streaming Ganglion  + BLED112 Dongle + USB C + Mac


TenTennyson:
## Problem

Slow data showing in GUI v5.1.0 on macOS 

## Expected 

Data will resume normal expectancy that is showed in Windows

## Operating System and Version

macOS Big Sur
Version 11.6
MacBook Pro (13-inch, M1, 2020)
Chip Apple M1
Memory 8GB

## GUI Version

GUI v5.1.0

## Running standalone app

Running the downloaded app

## Type of OpenBCI Board
Ganglion

## Are you using a WiFi Shield?

No

retiutut:
![Screen Shot 2022-09-22 at 11 04 13 AM](https://user-images.githubusercontent.com/25914123/191796778-dfaf28b9-e035-4cd6-9a81-e18c5275742f.png)

retiutut:
Does not work on this Mac
![Image 9-22-22 at 11 05 AM](https://user-images.githubusercontent.com/25914123/191797237-331291af-0e82-4751-978e-17f6be3d8a66.jpg)

retiutut:
Does work on this Mac
![BC2A6F74-637A-42EF-9C85-FB3F5B03E510](https://user-images.githubusercontent.com/25914123/191843192-f62abf8e-2103-4775-9c68-8342bd998af4.jpeg)

retiutut:
https://github.com/brainflow-dev/brainflow/issues/568

retiutut:
This issue occurs when plugging in the BLED112 Dongle (USB Type A) into a USB C Hub or a USB Type C to Type A converter. We see a slower sampling rate which ranges from 82Hz (worst case) to 195Hz (best case). The desired sampling rate is 200Hz. An issue has been logged at the BrainFlow level to see if we can go back to using the built-in bluetooth on Macs. This was possible with the now-deprecate OpenBCI Hub (GUI v4 and NodeJS). But now, Apple silicon exists and we have the BrainFlow backend. This issue has been fully investigated with quite a few USB C adapters and hubs and we believe this is the path forward to resolve this for Mac users.

wjcroft:
**Another perspective on the assumption that this is caused by ‘bad’ USB-C to USB-A adapters.**

[...]

PR #1090, opened 2022-12-07
Ganglion Native BLE

retiutut:
Fixes https://github.com/OpenBCI/OpenBCI_GUI/issues/1080


Andrey1994:
In brainflow everything is done, and its in the process of porting to the gui...

philippitts:
In later versions of MacOS (generally MacOS 11+) there is an issue where Bluetooth permissions are not requested by the application. This prevents the embedded native BLE library from accessing the Bluetooth adapter and therefore Ganglion boards (and other Bluetooth devices) cannot be found.

This issue appears to be related to [application hardening requirements](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution) introduced in MacOS 10.14.5. Most importantly, the JDK 8 version that ships with Processing 3.5.4 and earlier [is not notarized](https://github.com/adoptium/temurin-build/issues/1130) and therefore cannot request permissions from the OS.

A workaround to the issue is updating the JDK in Processing to the latest version of JDK 8. This can be done with the following:

```
curl -O -L --insecure https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u352-b08/OpenJDK8U-jdk_x64_mac_hotspot_8u352b08.tar.gz
tar -xf OpenJDK8U-jdk_x64_mac_hotspot_8u352b08.tar.gz
mv jdk8u352-b08 /Applications/Processing.app/Content/PlugIns/jdk1.8.0_202.jdk
```

Then follow the instructions for [running from the IDE](https://docs.openbci.com/Software/OpenBCISoftware/GUIDocs/#running-the-openbci-gui-from-the-processing-ide).

Note that applications exported from Processing 3.5.4 do not appear to ask for Bluetooth permissions even if the JDK is updated before exporting. Processing 3.5.4 is no longer maintained so downstream fixes may exist in Processing 4.

OpenBCI_GUI/ControlPanel.pde
```diff
@@ -1161,6 +1201,21 @@ class InterfaceBoxGanglion {
...
     }
 
+    private void createGanglionNativeBLEButton(String name, String text, boolean isToggled, int _x, int _y, int _w, int _h) {
+        protocolGanglionNativeBLE = createIFBGButton(name, text, isToggled, _x, _y, _w, _h);
+        protocolGanglionNativeBLE.onRelease(new CallbackListener() {
+            public void controlEvent(CallbackEvent theEvent) {
+                controlPanel.wifiBox.wifiList.items.clear();
+                controlPanel.bleBox.bleList.items.clear();
+                selectedProtocol = BoardProtocol.NATIVE_BLE;
+                //controlPanel.bleBox.refreshGanglionBLEList();
```
philippitts, line 1211: Is this needed? Recommend uncommenting it if it's useful or removing it if it's not.

OpenBCI_GUI/OpenBCI_GUI.pde
```diff
@@ -331,7 +332,7 @@ void settings() {
         win_w = 980;
         win_h = 580;
     }
-    size(win_w, win_h, P2D);
+    size(win_w, win_h);
```
philippitts, line 335: Why the renderer change here? If this is a leftover from exploring Processing 4 upgrade paths, we should save it for a Processing 4 specific PR.

philippitts (changes requested):

OpenBCI_GUI/OpenBCI_GUI.pde
retiutut: Correct. Will do.

OpenBCI_GUI/ControlPanel.pde
retiutut: I'll re-evaluate this line. Thanks. This PR needs a few more UX passes from myself, and I'll address these requested changes also.

philippitts (approved):

retiutut:
Wooo! Let's gooooo!

PR #1099, opened 2023-01-12
Networking LSL avgbandpower only one band value is sent

### LINKED ISSUE #1098: Networking LSL AvgBandPower: only one band value is sent
## Problem

When using the LSL networking protocol to send the AvgBandPower, the data output is only one single value.

## Expected 

Like for the other protocol, the expected output should be the 5 avg power value (one per band).
In the W_Networking.pde/sendNormalizedPowerBandData function,  the loop across each frequency band is missing for the LSL case.

Windows

## GUI Version
5.1.0

## Running standalone app
downloaded app

## Type of OpenBCI Board
Synthetic

No

## Console Log

Thank you!

retiutut:
Fixes #1098 

Screen Shot 2023-01-11 at 9 16 29 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,407 +3,473 @@
 ### Bug Fixes
 
 ### Improvements
-* Update to BrainFlow 5.1.0
+
+- Update to BrainFlow 5.1.0
+- Fix LSL AvgBandPower data type only one value is sent #1098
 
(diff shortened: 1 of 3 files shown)


retiutut:
It looks like Prettier markdown formatter has standardized the Changelog file.

PR #1120, opened 2023-02-08
Refactor the creation and playback of OpenBCI GUI data files

### LINKED ISSUE #1119: Refactor Playback Mode Classes
## Problem

Support #1113 and other internal OpenBCI work by completely refactoring DataSourcePlayback and DataWriterODF. This aligns with a recent BrainFlow change that supports a default preset for data and an Auxiliary preset for additional data channels. Sometimes, data could be saved into two separate files. 

## Expected 

Base DataSourcePlayback file, then extend this class for DataSourcePlaybackCyton, DataSourcePlaybackGanglion, etc.

All OS

## GUI Version

GUI 5.1.1 - development branch

VSCode

## Type of OpenBCI Board

All boards!!!

N/A

retiutut:
- Fixes #1119 

- Also changes Default Railed percentages to 90% and 75%.


retiutut:
Still TODO:

- [ ] Create playback files using every board class
- [ ] Test playback files from every board class

OpenBCI_GUI/BoardBrainFlowSynthetic.pde
```diff
@@ -2,13 +2,9 @@ import brainflow.*;
 
 import org.apache.commons.lang3.tuple.Pair;
 
-class BoardBrainFlowSynthetic extends BoardBrainFlow
-implements AccelerometerCapableBoard, PPGCapableBoard, EDACapableBoard, BatteryInfoCapableBoard {
+class BoardBrainFlowSynthetic extends BoardBrainFlow implements AccelerometerCapableBoard {
```
Andrey1994, line 5: why did you remove other interfaces? Synthetic boards still supports them..

Andrey1994 (commented):
I dont get why synthetic board for brainflow is changed, it still has all these datatypes and maybe they can be useful for somebody

OpenBCI_GUI/BoardBrainFlowSynthetic.pde
retiutut: I removed them because the widgets to visualize these data types do not exist in this copy of the GUI.

Andrey1994: But now we have interfaces like EDACapableBoard which are not implemented by any class and not used at all. And they are not removed as a part of this PR... if you want to make it consistent probably we need to remove them but it may be more complicated to sync two forks in that case. So probably it makes sense to restore them in synthetic board

retiutut: Good catch. All that is left in this repo are these classes. I will remove them.

OpenBCI_GUI/DataWriterODF.pde
retiutut, line 1: Fix line endings.

OpenBCI_GUI/FileBoard.pde
retiutut, line 1: Fix line endings

OpenBCI_GUI/SignalCheckThresholds.pde
retiutut, line 1: fix line endings

OpenBCI_GUI/W_Playback.pde
retiutut, line 1: Fix line endings

OpenBCI_GUI/W_PulseSensor.pde
retiutut, line 1: Fix line endings

retiutut (commented):
Fix line endings in a few files.

retiutut:
Tested with:
- Cyton
- Cyton Daisy
- Cyton Daisy + Accelerometer
- Cyton Daisy + Analog Read
- Cyton Daisy + Digital Read (with pressing PROG)
- Ganglion Native Bluetooth
- Ganglion BLE112 Dongle
- Synthetic Board (yes this generates a file for testing and development reasons)
- Playback History Widget playing back all of the above use cases in rapid succession within one session

retiutut:
In W_Playback the MenuList object does not perform well upon resizing. This should be replaced by an array of buttons or some other solution. All other calls to this UI object are not resized, as they are in the Control Panel.

retiutut:
Sending to NYC office for final testing as performed here: https://github.com/OpenBCI/OpenBCI_GUI/pull/1120#issuecomment-1536686587

isaacplotkin (changes requested):
Playback mode is working great on my end. I tested all the board configurations you mentioned. A few notes:
- The first time I click “Select OpenBCI playback file” after launching the GUI, the open file explorer window opens behind my other windows (Windows 10). Only happens the first time I click the button.
- Error when trying to record a 16 channel synthetic stream. Don’t get this problem when testing GUI 5.1.0
image
- Maybe add name of selected file below “Select OpenBCI playback file”. Not necessary but it confused me at first if I had selected a file

retiutut:
> Playback mode is working great on my end. I tested all the board configurations you mentioned. A few notes:
> 
> * The first time I click “Select OpenBCI playback file” after launching the GUI, the open file explorer window opens behind my other windows (Windows 10). Only happens the first time I click the button.
> * Error when trying to record a 16 channel synthetic stream. Don’t get this problem when testing GUI 5.1.0
>   image
> * Maybe add name of selected file below “Select OpenBCI playback file”. Not necessary but it confused me at first if I had selected a file

Great catch! I'll fix this here in this PR. The session settings for both playback and synthetic sessions were misnamed.

We do say in the status bar at the bottom of the GUI that the file was either successful or failed to load the file. Also appears in the console log. We do this in the status bar since there is enough space to display longer file names.

Screen Shot 2023-05-09 at 4 42 35 PM

retiutut:
Ok @isaacplotkin, the fix for synthetic mode not launching is in.

isaacplotkin (approved):

PR #1130, opened 2023-03-27
1129 fix time series x axis label not showing if last channel is hidden

retiutut:


retiutut:
Passing self code review and testing.

Issue #893, opened 2020-10-08
Look into Cyton SD Card Recording


retiutut:
## Problem

https://openbci.com/forum/index.php?p=/discussion/2552/recording-crashes-with-sd-card-and-other-issues#latest

This could be a firmware issue or maybe the wrong commands are sent to the board from the GUI. This is worth looking into to determine the root cause.

## Expected 

Record SD card files with maximum file size specified in Control Panel. Files should have no errors and playback fine in GUI v5.

## Operating System and Version

All OS

## GUI Version

GUI v5.0.0+

## Running standalone app

Are you running the downloaded app or are you running from Processing or Visual Studio Code? **Both**

## Type of OpenBCI Board

Cyton and Cyton+Daisy

## Are you using a WiFi Shield?

No

## Console Log

See Forum thread posted above for additional information.

retiutut:
https://openbci.com/forum/index.php?p=/discussion/3040/stabilize-sd-card-recording#latest

retiutut:
https://openbci.com/forum/index.php?p=/discussion/3150/question-on-format-not-matching-for-sd-card-log-as-described#latest

retiutut:
https://openbci.com/forum/index.php?p=/discussion/3235/sd-card-logging-interval-asked-for-not-reflected-in-recording#latest

retiutut:
https://openbci.com/forum/index.php?p=/discussion/3235/sd-card-logging-interval-asked-for-not-reflected-in-recording#latest

>Sure, here is the below logic which I have implemented for firmware change.

For 5 min, in firmware count of 11000 is mentioned, but currently this records for only 3 mins, so that means 3 mins represents 11000, so for 1 min it will be 11000/3, and for 5 min it will be (11000/3)*5 = 11000 *(5/3)

5 mins represents 3 mins ----> 11000 ----> then for 1 min ----> (11000/3) ----> for 5 mins ----> (11000/3)5 = (11000)(5/3)
15 mins represents 9 mins ----> 33000 ----> then for 1 min ----> (33000/9) ----> for 15 mins ----> (330000/9)15 = (33000)(5/3)
30 mins represents 18 mins ----> 66000 ----> then for 1 min ----> (66000/18)----> for 30 mins ----> (66000/18)30 = (66000)(5/3)

Same will continue for 15 mins, and other time intervals.


[...]

PR #1135, opened 2023-05-16
Fix CytonDaisy + SD Card sampling rate

retiutut:
CytonDaisy over-the-air sampling rate is 125, but it happens to store to SD card at 250. Tested and confirmed twice.

Relates to #893

### DIFF
--- a/OpenBCI_GUI/DataSourceSDCard.pde
+++ b/OpenBCI_GUI/DataSourceSDCard.pde
@@ -51,7 +51,7 @@ class DataSourceSDCard implements DataSource, FileBoard, AccelerometerCapableBoa
             }
             else {
                 if (samplingRate == 0) {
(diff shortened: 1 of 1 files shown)


retiutut:
##Testing

- Create a 1-3 minute recording using Cyton 8ch and SD card. Set to 5 minute maximum.
- Create another 1-3 minute recording (approximately same length as first recording) using CytonDaisy 16ch and SD card. Set 5 minute maximum.

- Playback both files using GUI 5.1.0. The CytonDaisy length should appear twice as long as it should be.
- Playback both files using `this` branch. The CytonDaisy length should appear as the correct recording length.

tharun-OpenBCI (approved):
SD card recording fixed for Cyton and Cyton+Daisy. File sizes match

PR #1173, opened 2023-08-01
Update Ganglion decompression

philippitts:


retiutut (commented):
Can we see some screenshot(s) of this working with the new firmware as a substitute for testing?

philippitts:
This is 5mV at 32Hz
image (1)

This is 5mV at 12
image (2)
Hz

This is 1mV at 19Hz
image (3)

This is 3mV at 7Hz
image (4)

This is 4mV at 4Hz
image (5)

And this is 4mV at 4Hz on the old firmware (just as a sanity check)
image (6)

.gitignore
```diff
@@ -26,3 +26,4 @@ libDataHandler.so
 libGanglionLib.so
 libGanglionScan.so
 libunicorn.so
+OpenBCI_GUI/out/
```
retiutut, line 29: This is fine, but curious as to what generates this?

philippitts: Building/running processing from VSCode created that folder for me.

retiutut: https://github.com/OpenBCI/OpenBCI_GUI/wiki/Developer-Setup#visual-studio-code-set-up

It should actually be building into `build` using the suggested settings.

retiutut: This is the one thing I would like changed.

retiutut (changes requested):

.gitignore
philippitts: Since it's the default build location, I think we should leave it in. Creating the task JSON file is not really necessary since you can just use Ctrl + Shift + P and hit enter to run processing. Ctrl + Shift + B also overrides other common build shortcuts from other extensions (e.g. C++). It's great if you want to configure it, but I think we should also support the default way without creating a custom shortcut in VSCode.

retiutut: I'm not seeing this as an option for me, maybe it was recently added to the VS Code extension?

PR #1174, opened 2023-08-02
Add Software Marker Widget

### LINKED ISSUE #1091: Add Software Marker Widget
Traditionally, OpenBCI users can use Cyton external input pins to add markers by pressing a button, photoresistor, or from an Arduino using an optoisolator. Ganglion does not have external pins. Some users want to add markers to the data stream from the software side.

Add a new widget to the OpenBCI GUI that allows users to press a keyboard button or mouse click to add a marker to the data stream. This should now be possible using the BrainFlow Markers feature.

This could help many users on Day 0 with hardware in their hands!

Cyton users have been using hardware/external triggers for years. Ganglion users may have designed custom software triggers using 3rd party software. Both options are still accessible, but maybe we can make something that can be distributed to all OpenBCI users.

## **Additional context**
Simple widget that accepts keyboard button press or digital button inside the widget with mouse click.

### LINKED ISSUE #1114: Allow markers to be set via network connection
When recording data from the GUI and streaming it to other applications, it is not possible to place markers in the data from the external application. This would be useful in many cases where events in the external application relate to events of interest and it is not possible to manually push a marker button in the GUI.

Extend the marker widget from https://github.com/OpenBCI/OpenBCI_GUI/issues/1091 to allow markers to be set via a network connection.

There is a marker channel in BrainFlow boards, but BrainFlow streaming boards are one-directional. It might be possible to implement features to support this use case in BrainFlow, but it would likely require a new or modified communication mechanism.

retiutut:
Closes #1091 
Closes #1114

Account for the following:
- [x] Insert markers using clickable buttons
- [x] Have multiple markers (e.g. 1.0, 2.0, etc)
- [x] Insert markers using keyboard shortcuts (even when widget is closed)
- [x] Print marker and timestamp to console log
- [x] Time series graph to show markers over time (live and playback)
- [x] Networking Widget output
- [x] UDP marker receiver and UI in the widget #1114 

____
Other consequential changes while reviewing keyboard shortcuts:
- remove `s` shortcut to Stop Streaming
- remove `b` shortcut to Start Streaming
- remove `d` shortcut as it was unused (it used to be for Cyton to reset board to defaults, this would not sync with ADS1299 settings UI)
All of the above shortcuts are no longer useful or could cause issues if accidently pressed. We can still start/stop streaming using spacebar.


retiutut:
image

Markers making their way into OpenBCI CSV files. Loaded into spreadsheet and sorted that column. You can see the markers next to the timestamp channel.

retiutut:
image


We will print an approximate time that a marker was inserted into the data. Here we see that the timestamp was correct down to the millisecond!

retiutut:
image

Markers showing up in playback!

retiutut:
LSL Networking Test
image

UDP Networking Test (everything is right but the sampling rate calculation is incorrect)
image

OSC Networking Test
image

philippitts (changes requested):
- [ ] Add some padding on the right side of the buttons
- [ ] Consider making the buttons larger and splitting them into two rows
- [ ] Autoscale the height of the graph

retiutut:
> * [ ]  Add some padding on the right side of the buttons
> * [x]  Consider making the buttons larger and splitting them into two rows
> * [ ]  Autoscale the height of the graph

I don't think autoscaling is the way to go here (for now), until users are allowed to stream in floats from another program. Then, it should be the only option for that source.

retiutut:
image
image


Successfully inserting markers (as floats) from external process over UDP.

retiutut:
image

Networking-Test-Kit/UDP/udp_receive_marker.py
philippitts, line 1: It looks like there is some commented out code that can be removed here.

Networking-Test-Kit/UDP/udp_send_marker.py
```diff
@@ -0,0 +1,42 @@
...
+
+    test_duration = 10
+
+    # Collect command line arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--ip", default="127.0.0.1",
+      help="The ip of the OSC server")
+    parser.add_argument("--port", type=int, default=12350,
+      help="The port the OSC server is sending to")
+    args = parser.parse_args()
```
philippitts, line 17: These arguments talk about an OSC server, but this is a UDP utility. I think this should be updated to UDP. It would also make sense to capitalize IP in this help stream.

OpenBCI_GUI/Board.pde
philippitts, line 1: What is actually changed here? Looks like line endings or something caused the whole file to be a diff.

OpenBCI_GUI/W_CytonImpedance.pde
philippitts, line 1: What actually changed here? Looks like the whole file is a diff.

philippitts (changes requested):
- [ ] It looks like there a few files where the whole file ended up as a diff for some reason. Can you double check your line endings and try restaging the files?
- [ ] I'm not sure how to provide external triggers via something like [netcat](https://netcat.sourceforge.net/). Can you provide some information (PR comment is fine) on this? Should the external process listen or host the connection? Does the stream have to be started first?

philippitts:
> > * [ ]   Add some padding on the right side of the buttons
> > * [x]   Consider making the buttons larger and splitting them into two rows
> > * [ ]   Autoscale the height of the graph
> 
> I don't think autoscaling is the way to go here (for now), until users are allowed to stream in floats from another program. Then, it should be the only option for that source.

Maybe we should add it now that it's possible to stream in floats from an external source?

retiutut:
> > > * [ ]    Add some padding on the right side of the buttons
> > > * [x]    Consider making the buttons larger and splitting them into two rows
> > > * [ ]    Autoscale the height of the graph
> > 
> > 
> > I don't think autoscaling is the way to go here (for now), until users are allowed to stream in floats from another program. Then, it should be the only option for that source.
> 
> Maybe we should add it now that it's possible to stream in floats from an external source?

Instead of this, users now have full control over the Y Axis, up to 20.0. Autoscale is kind of weird here since all values should be greater than 0, and it may cause "bouncing" when the graph resizes. Overall, the UX is likely better without the Y axis dynamically changing.

Networking-Test-Kit/UDP/udp_receive_marker.py
retiutut: Sure, these kinds of scripts usually get changed when debugging. Leaving just 1-2 lines in that help with that.

Networking-Test-Kit/UDP/udp_send_marker.py
retiutut: I agree. Thanks for helping clean this up since it's a new file.

retiutut:
> * [x]  It looks like there a few files where the whole file ended up as a diff for some reason. Can you double check your line endings and try restaging the files?
> * [ ]  I'm not sure how to provide external triggers via something like [netcat](https://netcat.sourceforge.net/). Can you provide some information (PR comment is fine) on this? Should the external process listen or host the connection? Does the stream have to be started first?

In the included Python example in `NetworkingTestScripts` folder, we send a single `float` over UDP to the specified IP and Port. Though, **the float is packed as a byte array**.

PR #1175, opened 2023-08-07
Fix brainflow library unpacking for MacOS

philippitts:


philippitts:
Looks good. Tested both running from Processing 4.3 and running the built DMG on MacOS 12.5.1.

philippitts:
We need to update the documentation to notify developers on MacOS that they will need to set the following environment variable when running from Processing.

```
export ld_library_path=/path/to/libraries/brainflow
```

philippitts:
We also need to update the license agreement popup when installing the application from the DMG on MacOS.

retiutut:
This looks similar to the fix for OpenBCI_GUI_Helpers library!

retiutut (approved):

retiutut (changes requested):
@philippitts Does this mean we need to copy/paste all dynamic libraries here again when updating brainflow?

retiutut (approved):

Issue #1094, opened 2023-01-05
Target widget must be open in order to stream from networking widget


philippitts:
## Problem

Data will not stream from the networking widget unless the widget that is the source of that data is open. For example, it is not possible to stream average band powers unless the band power widget is open.

## Expected 

Data will stream from the networking widget regardless if the widget that is the source of that data is open.

retiutut:
Drawing is the most expensive thing we do in Processing GUI. It shouldn't be that big of a hit to performance to compute all metrics in the background.

retiutut:
Networking Outputs, and those that need data processed in the background:

- Time Series filtered - Done
- Time Series unfiltered - Done
- **Focus - NEED**
- FFT - Done
- EMG - Done
**- AvgBandPower - NEED**
**- BandPower - NEED**
**- Accel/Aux - NEED**
**- Pulse - NEED**
**- EMG Joystick - NEED**
- Marker - Done

philippitts:
Closed by https://github.com/OpenBCI/OpenBCI_GUI/pull/1181

PR #1181, opened 2023-08-28
Process widget data in background when widgets are closed

retiutut:
Networking Outputs, and those that need data processed in the background:

- [x] Time Series filtered - Done
- [x] Time Series unfiltered - Done
- [x] **Focus - NEED**
- [x] FFT - Done
- [x] EMG - Done
- [x] **AvgBandPower - NEED**
- [x]  BandPower - Done
- [x] Accelerometer - NEED
- [x] Digital Read - NEED
- [x] Analog Read - NEED
- [x] **Pulse - NEED**
- [x] **EMG Joystick - NEED**
- [x] Marker - Done
- [x] Process sampling rate dependent data accumulation in background when Networking widget is closed


OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+                    println(e.getMessage());
+                }
+            } else {
+                // This method has been updated to reduce duplicate packets - RW 3/15/23
+                if (checkForData()) {
+                    sendData();
+                }
+            }
+        }
+    }
```
philippitts, line 171: Optional - you might consider extracting the LSL and non-LSL logic into separate methods.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+                    sendDigitalReadData();
+                }
+            }
+        } else if (this.dataType.equals("Pulse")) {
+            sendPulseData();
+        } else if (this.dataType.equals("EMGJoystick")) {
+            sendEMGJoystickData();
+        } else if (this.dataType.equals("Marker")) {
+            sendMarkerData();
+        }
```
philippitts, line 261: Optional - a switch statement may be more clear here, offer better logical flow, and reduce the number of comparisons required. The number of comparisons matters since this code runs in an update loop.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+            float sampleRateCheck = samplesSent / timeDelta;
+            println("\nNumber of samples collected = " + samplesSent);
+            println("Time Interval (Desired) = " + (sampleRateClockInterval / 1000));
+            println("Time Interval (Actual) = " + timeDelta);
+            println("Sample Rate (Desired) = " + currentBoard.getSampleRate());
+            println("Sample Rate (Actual) = " + sampleRateCheck);
+            sampleRateClock = 0;
+            samplesSent = 0;
+        }
+        */
```
philippitts, line 299: Optional - it might make sense to extract this to a separate method and then have a single call to that method commented out here.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+                    println(e.getMessage());
+                }
+            }
+
+        }
+    }
+
+    // Send out 1 or 0 as an integer over all networking data types for "Focus" data
+    private void sendFocusData() {
+        final int IS_METRIC = w_focus.getMetricExceedsThreshold();
```
philippitts, line 387: Action - this name is confusing. MetricValue or similar would be better. I recommend not prefixing with 'Is' since it's not a boolean. The name should also be more about the value than the fact it is a metric (it's always a metric regardless if it's currently exceeding the threshold).

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+                * message for (int j=0;j<125;j++) { float fft_band = fftBuff[i].getBand(j);
+                * String fft_band_3dec = threeDecimalPlaces.format(fft_band); serialMessage +=
+                * fft_band_3dec; if (j < 125-1) { serialMessage += ","; //add a comma to
+                * serialMessage to separate chan values, as long as it isn't last value... } }
+                * serialMessage += "]"; try { // println(serialMessage);
+                * this.serial_networking.write(serialMessage); } catch (Exception e) {
+                * println(e.getMessage()); } }
+                */
+        }
+    }
```
philippitts, line 486: Optional - For all of these send methods, it would read more clearly if the individual network type sections had their own send methods. For example:

```
private void sendFFTData() {
    if (this.protocol.equals("OSC")) sendFFTDataOSC();
    else if (this.protocol.equals("UDP")) sendFFTDataUDP();
    else if (this.protocol.equals("LSL")) sendFFTDataLSL();
    else if (this.protocol.equals("Serial")) sendFFTDataSerial();
}
```

You could even do the protocol string comparison in `sendData` and then pass it down as an enum to speed things up. I wouldn't recommend using a switch statement here because it would decrease the readability for some minor speed increases, but I could see it either way.

OpenBCI_GUI/NetworkStreamOut.pde
philippitts: Even better - use a global static final `DEBUG` variable or similar and execute the method if it is set.

philippitts: There are several other commented out blocks that would benefit in the same way.

philippitts: Even better - create a logger class and log information there at different severity levels. This will allow you to print, save, or send the debug output however you like and give you one section of code to manage logging.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+        // UNFILTERED & FILTERED ... influenced globally by the FFT filters dropdown
+        // just like the FFT data
+        final int NUM_BAND_POWERS = 5; // DELTA, THETA, ALPHA, BETA, GAMMA
+
+        if (this.protocol.equals("OSC")) {
+            for (int i = 0; i < numExgChannels; i++) {
+                msg.clearArguments();
+                msg.setAddrPattern(baseOscAddress + "/band-power/" + i);
+                for (int j = 0; j < NUM_BAND_POWERS; j++) {
+                    msg.add(dataProcessing.avgPowerInBins[i][j]); // [CHAN][BAND]
```
philippitts, line 498: Action - Why not just name them channel and band?

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+    private void openNetwork() {
+        println("Networking: " + getAttributes());
+        if (this.protocol.equals("OSC")) {
+            // Possibly enter a nice custom exception here
+            // try {
+            this.osc = new OscP5(this, this.port + 1000);
+            this.oscNetAddress = new NetAddress(this.ip, this.port);
+            this.msg = new OscMessage(this.baseOscAddress);
+            // } catch (Exception e) {
+            // }
```
philippitts, line 1211: Action - This code should be removed or the exception should be implemented.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1262 @@
...
+            this.udp.listen(false);
+            this.udp.log(false);
+            output("UDP successfully connected");
+        } else if (this.protocol.equals("LSL")) {
+            String stream_id = "openbcigui";
+            info_data = new LSL.StreamInfo(this.streamName, this.streamType, this.numLslDataPoints,
+                    currentBoard.getSampleRate(), LSL.ChannelFormat.float32, stream_id);
+            outlet_data = new LSL.StreamOutlet(info_data);
+        } else if (this.protocol.equals("Serial")) {
+            // Open Serial Port! %%%%%
```
philippitts, line 1224: Info - what does `%%%%` mean?

philippitts (changes requested):

philippitts:
Tested and confirmed the changes are working as expected 👍

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: It actually is a boolean, though represented as a 0 or 1.

retiutut: Changing to `metricValue`. This should not have been IS_METRIC since it's a local final variable. I also agree with word choice as you have said.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: Agreed.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: Yes.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: I have no clue.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: Though I don't like all of the `if` statements in this class, keeping the sendDatatypeData() method together does help with development when comparing how data is sent over multiple protocols.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: Putting this as it's own method since it's reused and hiding behind a debug flag.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1272 @@
...
+            // Filtered
+            udpDataTypeName = "timeSeriesFilt";
+            oscDataTypeName = "time-series-filtered";
+            for (int i = 0; i < newDataFromBuffer.length; i++) {
+                newDataFromBuffer[i] = w_networking.dataBufferToSend_Filtered[i];
+            }
+        }
+
+        debugTimeSeriesDataSamplingRate();
+
```
philippitts, line 327: Action - It would be better to check the `debugSamplingRate` flag here rather than in the method. The method shouldn't be called if the flag isn't set.

OpenBCI_GUI/NetworkStreamOut.pde
```diff
@@ -0,0 +1,1272 @@
...
+    private void sendMarkerData() {
+
+        float[] newDataFromBuffer = new float[nPointsPerUpdate];
+
+        for (int i = 0; i < newDataFromBuffer.length; i++) {
+            newDataFromBuffer[i] = w_networking.markerDataBufferToSend[i];
+        }
+
+        debugTimeSeriesDataSamplingRate();
+
```
philippitts, line 1123: Action - same as above. Check the debug flag here rather than in the function. If the function is called, the debugging should be done.

philippitts (approved):
I recommend checking the debug flag before entering the debug method, rather than after. The function name should be something of a contract to the developer. If it says it will do something, it should do it. In this case, the function name is `debugTimeSeriesDataSamplingRate` but in truth it's more like `debugTimeSeriesDataSamplingRateIfDebugFlagIsSet` as implemented. That could be confusing to a future developer who calls that function, but doesn't set the flag. Setting the flag outside the function avoids that confusion.

Everything else looks good - no need to re-request a review after fixing the above. I'll approve now and you can go ahead and merge after.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: Will do.

OpenBCI_GUI/NetworkStreamOut.pde
retiutut: Done.

retiutut:
Closes #1094

PR #1186, opened 2023-09-12
Add EV certificate signing for Windows builds

philippitts:
- Replaces code signing for Windows builds with Extended Validation code signing which provides additional guarantees to end-user systems and reduces warning about untrusted software
- Extracts code signing along with other OS specific logic from the `make_release.py` script
- Decomposes the `make_release.py` script into a `build.py` and `package.py` script


retiutut:
Looking great so far. Did an early review.

philippitts:
@retiutut can you double check that the BrainFlow library setup for MacOS is still correct? I ran builds on Windows and MacOS and they seem correct, but it would be great to have you confirm it.

philippitts:
The v5 GUI used to store build artifacts with timestamps in the names, but we've taken those out for v6. If you prefer to have timestamps in the name we can just drop the related commit and it should work. However, I think we can use versioning for the artifacts in AWS and it will work better.

release/package.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Write the file out again
+        with open(build_directory + '/OpenBCI_GUI', 'w') as file:
+            file.write(filedata)
+
+        print ("Fixed issue on Linux when launching from directory with spaces.")
+
+    if LOCAL_OS == MAC:
+        shutil.move(flavors[LOCAL_OS] + ".dmg", new_name + "macosx.dmg")
+    else:
```
retiutut, line 63: It's not explicit here that this is the case for Windows. Can we make it more obvious?

retiutut (changes requested):

retiutut (changes requested):

philippitts:
Updated the PR based on your feedback and tested on Windows, MacOS, and Linux. Things are working as expected and it's ready for another review pass.

PR #1196, opened 2023-09-28
Fix GUI v6.0.0-beta.0 crash when resizing during session

retiutut:


retiutut:
Fixes the issue and tested on macOS 14!

PR #1199, opened 2023-10-12
Add Focus Widget values to Session Settings - Fixes #1198

### LINKED ISSUE #1198: Add Focus Widget values to Session Settings
## Problem

Just noticed that this was not added to Session Settings when re-implemented.

## Expected 

Save/Load/Default dropdowns for the Focus Widget

All OS

## GUI Version

v6.0.0-beta.0

Both

## Type of OpenBCI Board

All boards and modes

```
    private final String kJSONKeyDataInfo = "dataInfo";
    private final String kJSONKeyTimeSeries = "timeSeries";
    private final String kJSONKeySettings = "settings";
    private final String kJSONKeyFFT = "fft";
    private final String kJSONKeyAccel = "accelerometer";
    private final String kJSONKeyNetworking = "networking";
    private final String kJSONKeyHeadplot = "headplot";
    private final String kJSONKeyBandPower = "bandPower";
    private final String kJSONKeyWidget = "widget";
    private final String kJSONKeyVersion = "version";
    private final String kJSONKeySpectrogram = "spectrogram";
    private final String kJSONKeyEmg = "emg";
    private final String kJSONKeyEmgJoystick = "emgJoystick";
    private final String kJSONKeyMarker = "marker";
 ```
 
 ^^^No key exists for Focus Widget and the values are private in the class 👎

retiutut:


retiutut:
CI/CD not configured properly for AWS credentials. Will fix. Good thing to find out in this tiny PR.

philippitts (changes requested):
When I try to save settings a couple issues occur:

1. The popup window for selecting the file location opens minimized. The GUI window also minimizes.
2. The default settings name is "SynthEightUserSettings" which doesn't make sense to me. What does this name mean?
3. When I save using the default "SynthEightUserSettings" name the program crashes.

If I load the settings first, and then try and save new settings it appears to work and none of the above issues occur. However, when I clear the settings and try to reload the newly saved settings, they do not load.

I noticed that three issues I listed at the top of this comment only occur if the Settings folder has not been created. I think the folder is created when loading settings which then allows subsequent saves to work.

The focus widget settings do not load for me regardless if the Settings folder has been created.

Running on Windows 10 from VSCode.

philippitts:
Updates:

* I'm having trouble reproducing the crash when saving the settings file. I'm not running as administrator so it could be related to that. When the AWS credentials are fixed I'll try running the built artifact as administrator to see if I can replicate the problem.
* Focus settings are loading. See my UX notes below.

Remaining issues:

* I just realized that "Synth" is "Synthetic" as applied to the user settings file name. I recommend spelling out the full word instead of using the abbreviation. Maybe I'm special, but it had me confused.
* When I click "Clear All" it deletes my files. Perhaps that is intended, but as a naive user of this feature I was surprised that it deleted my saved files and kept loading the default "SynthEightDefaultSettings" thinking it was my user saved settings (and then wondering why my settings didn't load).
* Sometimes when I load or save settings files the file selection window appears minimized. This may be because I'm not running as administrator and I'll re-test as described above.

Issue #1014, opened 2021-11-27
SD Card Hex to Standard Text File Conversion script


retiutut:
## **Is your feature request related to a problem? Please describe.**
Some users would appreciate an SD card Hex to human-readable text file for Cyton created files.

Maybe the files should look like the OpenBCI CSV files exactly, mimicking the header information. Something in the script would also need to determine if a Cyton or Cyton Daisy was used.

## **Describe the solution you'd like**
Do this in Python. Shouldn't be that complicated or difficult to run.

## **Describe alternatives you've considered**
Current GUI reads straight from Hex values. Though, some users would like to maybe have a human-readable CSV file that can be used in the GUI or with another program. 

There used to be a conversion script in GUI 4.2.0, maybe we should not point users to this, since the data CSV format changed from GUI v4 to GUI v5

## **Additional context**

https://openbci.com/forum/index.php?p=/discussion/1653/script-to-convert-sd-card-file#latest

retiutut:
https://openbci.com/forum/index.php?p=/discussion/3497/sdk-commands-to-get-the-data-files-from-sd-card-to-the-computer#latest

Another request for this feature. Upping the priority.

Shyamal-Dharia:
Hi, if possible I wanted to request you to prioritize this issue. 
Thanks

wjcroft:
So to summarize the current November 2024 status. User [roflecopter](https://github.com/roflecopter) has created a Python conversion script, done in January 2024.

https://github.com/roflecopter/openbci-session
https://github.com/roflecopter/openbci-session/blob/main/sd_convert.py

And user @Shyamal-Dharia has a GUI / Processing PR that will add a sd file convert feature to the GUI. Also completed and tested January 2024.

https://github.com/OpenBCI/OpenBCI_GUI/pull/1207

I believe both these conversion features should be made available to Cyton users. What is the next step? @retiutut @philippitts 

Regards,

PR #1207, opened 2024-01-20
added sd-card-file-conversion script

Shyamal-Dharia:
## Issue #1014 SD Card Hex to Standard Text File Conversion script.

Create a python script for sd card hex to csv file. additionally, it can also convert csv files to other formats using mne-python.

I have tested it with my sd card files and compared it with them. It works!

Also, couple of my files had analog values from photodiode sensor. It also converts those values and puts it in the last three column. Values can be distinguished by looking at graph generated from csv file from the column that had sensor values. The values ranged between 0 to 1. attached is the picture. Therefore, I think the script works for analog sensors too. 

Took reference from https://github.com/roflecopter/openbci-psg for conversion.

Hope this helps! 

graph

### DIFF
--- a/sd_file_conversion/txt_to_csv_conversion.py
+++ b/sd_file_conversion/txt_to_csv_conversion.py
@@ -1,5 +1,4 @@
 #OPENBCI HEX TO CSV CONVERSION REFERENCE: https://github.com/roflecopter/openbci-psg
-# GIVES output same as OpenBCI GUI 4.2
 
 import numpy as np
 import pandas as pd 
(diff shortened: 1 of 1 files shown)


retiutut:
GUI v4 files are very different from what will be GUI v6 files. Can you please download GUI v6 (beta) and try another pass at this? Awesome work so far!

retiutut (changes requested):
Suggest another pass to make compatible with GUI v6.

Shyamal-Dharia:
Understood. Yes, working on it will update the code as soon as possible.

Shyamal-Dharia:
The recent code update has been tested under the following scenarios with cyton+daisy:

1. With analog sensor values.
2. In digital mode, where the sensor was not attached.
3. With normal accelerometer data.

Based on my understanding, due to some limitations, the data for the following columns are not stored on the SD card: "Timestamp," "Marker Channel," and "Timestamp (Formatted)." Therefore, I was not able to use them and kept their values as zeros.

file type conversion also been tested for all three file types (.eeg, .edf, .set).

retiutut (approved):

retiutut (approved):

Issue #1058, opened 2022-05-12
Check Cyton Impedance on both N and P pins


retiutut:
## **Is your feature request related to a problem? Please describe.**
A user has requested that we be able to check the impedance on both Cyton pins.

## **Describe the solution you'd like**
Add the functionality to the Cyton Signal Widget to check both pins.  Maybe this should be hidden behind "expert mode". It's only useful when connecting electrodes to both the top and bottom pins for a channel. New users may not understand why we would check both pins if they are doing EEG.

The space already exists in the UI. See attached screenshot.

## **Describe alternatives you've considered**
We excluded this functionality initially since no team or community members saw value in this feature or requested it. Not very useful when doing EEG recordings, which require one electrode on the N pin only per channel.

## **Additional context**
Screenshot for reference.

![Screen Shot 2022-05-12 at 12 30 52 PM](https://user-images.githubusercontent.com/25914123/168134482-89a8dd99-7c7f-4a35-8ba7-642ed5555941.png)

marles77:
Looking forward to this feature in the next version. Sometimes I need to check my EOG electrodes as well.

PR #1211, opened 2024-02-29
Fix Cyton impedance check board configuration

### LINKED ISSUE #1204: Cyton impedance checks incorrect at low resistances
## Problem

Several users on the [OpenBCI forum](https://openbci.com/forum/index.php?p=/discussion/3726/electrode-impedance-test-using-the-gui-widget) report incorrect values from the electrode impedance widget for Cyton when using resistors of 500kOhms or less. Using resistors of greater than 500kOhms appear to be working as expected.

## Expected 

The Cyton impedance check widget should reflect the values of resistors used in lab validation configurations. If a resistor is connected between the N port of a channel and SRB2, the widget should show an impedance similar to the value of the resistor.

philippitts:
Closes #1204 

The Cyton impedance check configuration appears to be incorrectly disconnecting SRB2 and using Bias instead. This has led to incorrect impedance measurements as shown in https://github.com/OpenBCI/OpenBCI_GUI/issues/1204. This fix results in correct measurements with a 51 kOhm resistor - more tests to follow.

### DIFF
--- a/OpenBCI_GUI/BoardCyton.pde
+++ b/OpenBCI_GUI/BoardCyton.pde
@@ -450,8 +450,8 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
             currentADS1299Settings.values.gain[channel] = Gain.X1;
             currentADS1299Settings.values.inputType[channel] = InputType.NORMAL;
-            currentADS1299Settings.values.bias[channel] = Bias.INCLUDE;
-            currentADS1299Settings.values.srb2[channel] = Srb2.DISCONNECT;
+            currentADS1299Settings.values.bias[channel] = Bias.NO_INCLUDE;
+            currentADS1299Settings.values.srb2[channel] = Srb2.CONNECT;
             currentADS1299Settings.values.srb1[channel] = Srb1.DISCONNECT;
 
             fullCommand.append(currentADS1299Settings.getValuesString(channel, currentADS1299Settings.values));


philippitts:
| Resistor Value | Pin | GUI Impedance Reading |
| --------------- | --- | --------------------------- |
| 51 kOhm | SRB2 | 50 kOhm |
| 51 kOhm | 1P | 50 kOhm |

| Resistor Value | Pin | GUI Impedance Reading |
| --------------- | --- | --------------------------- |
| 470 kOhm | SRB2 | 550 kOhm |
| 470 kOhm | 1P | 553 kOhm |

| Resistor Value | Pin | GUI Impedance Reading |
| --------------- | --- | --------------------------- |
| 680 kOhm | SRB2 | 649 kOhm |
| 680 kOhm | 1P | 647 kOhm |

| Resistor Value | Pin | GUI Impedance Reading |
| --------------- | --- | --------------------------- |
| 1000 kOhm | SRB2 | 906 kOhm |
| 1000 kOhm | 1P | 902 kOhm |

retiutut:
It looks like the macOS build pipeline is broken. Though, I'm sure that this change is fine.

I'd like to replicate these results using resistors also as part of the review. The code is fine.

wjcroft:
Forum users have been reporting broken impedance function well before March 2024. I would estimate many months. Totally appreciate Philip tracking this down. 

On the Forum thread below I have suggested that users who need a workaround can try building the GUI from source using Phil's 2 line fix.

https://openbci.com/forum/index.php?p=/discussion/3790/high-impedance#latest
https://openbci.com/forum/index.php?p=/discussion/comment/20284/#Comment_20284

> Until the new GUI app download binaries are regenerated, you can apply the fix yourself using the GUI source code.
> 
> https://github.com/OpenBCI/OpenBCI_GUI/pull/1211
> https://github.com/OpenBCI/OpenBCI_GUI/pull/1211/commits/b3918c8814e1d7869e0c9430812724671bfda3ec
> https://github.com/OpenBCI/OpenBCI_GUI/issues/1204#issuecomment-1970392296
> 
> Currently most staff are busy on the Galea project, several important demos at conferences are happening this week. Once things settle back to normal, new download binaries should be available.
> 
> https://docs.openbci.com/Software/OpenBCISoftware/GUIDocs/#running-the-openbci-gui-from-the-processing-ide

On the other hand, MANY groups using Cyton for university research projects, depend on accurate impedance checks to comply with their academic research protocols. So a wider distribution of this repaired GUI would be welcome.

@jartuso also commented recently on Slack: "this is the cyton impedance problem that we've talked about before. @philippitts didn't we fix this? This one scares me." What you may want to be concerned about is that universities getting wind of this in larger numbers, could dampen sales.

OpenBCI_GUI/BoardCyton.pde
```diff
@@ -450,8 +450,8 @@ implements ImpedanceSettingsBoard, AccelerometerCapableBoard, AnalogCapableBoard
 
             currentADS1299Settings.values.gain[channel] = Gain.X1;
             currentADS1299Settings.values.inputType[channel] = InputType.NORMAL;
-            currentADS1299Settings.values.bias[channel] = Bias.INCLUDE;
-            currentADS1299Settings.values.srb2[channel] = Srb2.DISCONNECT;
+            currentADS1299Settings.values.bias[channel] = Bias.NO_INCLUDE;
+            currentADS1299Settings.values.srb2[channel] = Srb2.CONNECT;
```
retiutut, line 454: https://docs.openbci.com/GettingStarted/Biosensing-Setups/EMGSetup/#obtain-emg-data-with-openbci-cyton-board

```
Streaming EMG Data with the OpenBCI GUI

Once you have the GUI open, turn off any channels that you have not connected to muscle sensors.

Then for Cyton, go to hardware settings and turn off SRB2 for all of the channels that you are streaming data from.

Note: This is because ordinarily SRB2 is the reference point from which potentials are measured. However we are now having muscles reference themselves, so we do not want it to look at this pin.
```

While this may work, it also seems like this may break checking impedances when doing EMG.

retiutut: Relates to https://github.com/OpenBCI/OpenBCI_GUI/issues/1058

philippitts: It does not matter if you are reading EMG or EEG. The impedance check remains the same and measures the impedance of one Cyton electrode against SRB2, regardless of where that electrode is placed. Per the tests above, BIAS should never be used for impedance checks. This makes sense because BIAS is used as input to the ADS1299 to create destructive interference noise patterns that reduce noise vs SRB2 which is the reference point against which a channel can be measured (usually for EEG). If you look at my readings from the linked issue you can see that on almost all the BIAS tests the signal was steadily dropping. This is because the ADS1299 is performing exactly as expected and canceling out the "common noise" it's reading from the BIAS pin.

EMG doesn't use SRB2 while acquiring signals because EMG references against a second EMG channel (in most cases), but to measure impedance, SRB2 must still be used. The easiest method is to switch the BIAS electrode to SRB2 for impedance measurements and switch it back to BIAS for signal acquisition.

isaacplotkin:
I just validated Phil's results. Looks good to merge @retiutut

isaacplotkin:
| Resistor Value | Pin | GUI Impedance Reading |
| --------------- | --- | --------------------------- |
| 1000 kOhm | SRB2 | 966 kOhm |
| 1000 kOhm | 1P | 955 kOhm |

| Resistor Value | Pin | GUI Impedance Reading |
| --------------- | --- | --------------------------- |
| 100 kOhm | SRB2 | 100 kOhm |
| 100 kOhm | 1P | 100 kOhm |

tharun-OpenBCI:
Edit: Formatting

Retested this branch with Windows and GUI built through VSCode using Resistors of Values 330kΩ, 2MΩ and 5.6MΩ to simulate different thresholds of possible impedance values. 

It is to be noted that the ADS1299 current source used in impedance detection has an error of 20%. Due to the nature of impedance, this error is carried over to the Impedance values. 

Therefore, below are the possible ranges that the ADS1299 will measure for the provided resistor value:

1. 330kΩ - **264kΩ-396kΩ**
2. 2MΩ - **1.6MΩ-2.4MΩ**
3. 5.6MΩ - **4.48MΩ-6.72MΩ**

The values recorded are:

**330kΩ:**
| Channel Number | GUI Reading |
|--------------------|---------------|
| 1                          | 332kΩ       |
| 2                          | 330kΩ       |
| 3                          | 349kΩ       |
| 4                          | 385kΩ       |
| 5                          | 367kΩ       |
| 6                          | 342kΩ       |
| 7                          | 358kΩ       |
| 8                          | 335kΩ       |

**2MΩ:**
| Channel Number | GUI Reading |
|--------------------|---------------|
| 1                          | 1849kΩ       |
| 2                          | 1839kΩ       |
| 3                          | 1942kΩ       |
| 4                          | 1909kΩ       |
| 5                          | 1986kΩ       |
| 6                          | 1990kΩ       |
| 7                          | 2090kΩ       |
| 8                          | 1820kΩ       |

**5.6MΩ:**
| Channel Number | GUI Reading |
|--------------------|---------------|
| 1                          | 3823kΩ       |
| 2                          | 3770kΩ       |
| 3                          | 4047kΩ       |
| 4                          | 3881kΩ       |
| 5                          | 4139kΩ       |
| 6                          | 4106kΩ       |
| 7                          | 4386kΩ       |
| 8                          | 3709kΩ       |

Only the N inputs are accessible from the Cyton Signal Widget and this GUI change sets SRB2 for the electrode to be compared against. Since the appropriate P input is connected to the SRB2 pin internally in the ADS1299, measuring against P and SRB2 are the same although the value will be much more accurate if SRB2 is used. The above recorded values used the SRB2 pin.

tharun-OpenBCI (approved):

starsGu:
Why not test with resistances in the 5–10 kΩ range, which are closer to typical real-world EEG electrode impedances?

PR #1214, opened 2024-04-18
Fix detection of plugdev group for Linux users

MaKaNu:
- Add method isInElevatedGroup
- check if the linux user is in plugdev
- add logic for admin error check

Our users are always for root password. This is no Option.
We only use Cyton Dongle and this Improvment aims for Dongle usage.

To use Dongle under linux users must be in plugdev group.

### DIFF
--- a/OpenBCI_GUI/Extras.pde
+++ b/OpenBCI_GUI/Extras.pde
@@ -159,6 +159,34 @@ public boolean isElevationNeeded(String path) {
     return result;
 }
 /**
+* Determines if user is in elevated group (plugdev).  
+*
+* @return <code>true</code> if user is in group plugdev, <code>false</code> otherwise.
+ */
+public boolean isInElevatedGroup() {
+    boolean result = true;
+    if (isLinux()) {
+        try {
+            String command = "groups";
+            Process p = Runtime.getRuntime().exec(command);
+            p.waitFor();
+            InputStream stdIn = p.getInputStream();
+            BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(stdIn));
+            String[] values = bufferedReader.readLine().split(" ");
+            for (int idx=0; idx<values.length; idx++) {
+                result = values[idx].equals("plugdev");
+                if (result) {
+                    break;
+                }
+            }
+        } catch (Exception e) {
+            return false;
+        }
+    }
+    return result;
+}
+
+/**
 * Determine if user has administrative privileges.
 *
 * @return
--- a/OpenBCI_GUI/OpenBCI_GUI.pde
+++ b/OpenBCI_GUI/OpenBCI_GUI.pde
@@ -497,8 +497,8 @@ void delayedSetup() {
     //Apply GUI-wide settings to front end at the end of setup
     guiSettings.applySettings();
 
-    if (!isAdminUser() || isElevationNeeded()) {
-        outputError("OpenBCI_GUI: This application is not being run with Administrator access. This could limit the ability to connect to devices or read/write files.");
+    if (!isInElevatedGroup() && (!isAdminUser() || isElevationNeeded())) {
+        outputError("OpenBCI_GUI: This application is not being run with Administrator access or user is not member of plugdev group. This could limit the ability to connect to devices or read/write files.");
     }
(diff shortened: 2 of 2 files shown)


MaKaNu:
Can I get some response to this PR?

OpenBCI_GUI/Extras.pde
```diff
@@ -158,6 +158,34 @@ public boolean isElevationNeeded(String path) {
...
     return result;
 }
+/**
+* Determines if user is in elevated group (plugdev).  
+*
+* @return <code>true</code> if user is in group plugdev, <code>false</code> otherwise.
+ */
+public boolean isInElevatedGroup() {
+    boolean result = true;
+    if (isLinux()) {
```
retiutut, line 168: Reverse if and do if(!isLinux) return at the beginning of this method.

OpenBCI_GUI/Extras.pde
```diff
@@ -158,6 +158,34 @@ public boolean isElevationNeeded(String path) {
     }
     return result;
 }
+/**
+* Determines if user is in elevated group (plugdev).  
+*
+* @return <code>true</code> if user is in group plugdev, <code>false</code> otherwise.
+ */
+public boolean isInElevatedGroup() {
```
retiutut, line 166: Rename isInElevatedGroupLinux()

retiutut (changes requested):
Hello, work on GUI v7 has begun, and I would like to include this PR.

Let's keep this change isolated to only check on Linux and separated from other admin privileges checks.

retiutut (commented):
Also, can you please confirm this is a positive fix for **all Linux users**?

MaKaNu:
> Also, can you please confirm this is a positive fix for **all Linux users**?

What do you mean by **all Linux users**? Do you talking about adding unit tests? If that's the case, I am wiling to contribute some as well, but I am not sure how to simulate/mock the scenario.

retiutut:
> > Also, can you please confirm this is a positive fix for **all Linux users**?
> 
> What do you mean by **all Linux users**? Do you talking about adding unit tests? If that's the case, I am wiling to contribute some as well, but I am not sure how to simulate/mock the scenario.

I mean all Linux users, not just for your use case.

MaKaNu:
> I mean all Linux users, not just for your use case.

I just realized, that's mostly just a debian/ubuntu thing these days, to control usb access via static plugdev group. Further, it seems most other Distributions (Fedora, OpenSuse, Arch, even Gentoo) relay on `udev` and `uaccess`.

We also only tested for Cyton Biosensing Board Dongle, based on our availability.

In General, I would say yes, every Linux user is getting profit from changes in this area, since running GUI Applications in with sudo is in general a bad idea. Specific for us: We are dividing user and admin strictly on the systems, which makes sudo not accessible for the users.

An even better solution, would be proper packaging of the software, which also provide the specific udev rule with access to the usb dongle via uaccess.

retiutut:
So how does what you've described overlap/interact with an existing fix? Did you try the existing fix and it didn't work?

Found here: https://docs.openbci.com/Software/OpenBCISoftware/GUIDocs/#linux-users-serial-port-permissions

Side Note: We can realistically only target the most popular Linux distributions.

MaKaNu:
I just checked https://wiki.debian.org/SystemGroups And I can no longer remember, why we ended with the `plugdev` group. On the other Hand, the base of my MR is basic the reference of the fix you pointed against.

> Side Note: We can realistically only target the most popular Linux distributions.

Absolutly. I was just surprised, that only debian based distributions are the only one which are focusing on the group approach to apply udev rules. The uaccess approach seems to be the new way and debian is just the last horse.

I see two solution for the MR:

1. Implement it as the user information of https://docs.openbci.com/Software/OpenBCISoftware/GUIDocs/#linux-users-serial-port-permissions if the user is not dialup.
2. Going the packaging (debian, redhat) way and providing specific udev rules for the devices.

Or a combination of both.

Or (don't prefered by me) close it and rely on the docs.

retiutut:
Moving forward on this ticket, it should be separated out for Linux users and not combined with the Windows methods. See existing feedback.

PR #1247, opened 2025-04-27
Change BrainFlowStreamer recording to NONE by default

bleugreen:
Currently, the `DEFAULT` setting for the BrainFlowStreamer writes a .csv containing all session data (in addition to the .txt recorded by the GUI).

This means that unless it is manually set to NONE each time, there are two plaintext copies recorded per session, totaling ~8.5 MB/min. (~4.3 MB txt, ~4.2 MB csv)

This PR simply changes the initial BrainFlowStreamer enum value to NONE, so that only one copy is stored by default, while still allowing users to record data via BrainFlow if they prefer.

### DIFF
--- a/OpenBCI_GUI/ControlPanel.pde
+++ b/OpenBCI_GUI/ControlPanel.pde
@@ -1611,7 +1611,7 @@ class BrainFlowStreamerBox {
     private Button outputToNetwork;
     private Button outputToFile;
     private ScrollableList bfFileSaveOption;
-    private DataWriterBFEnum dataWriterBfEnum = DataWriterBFEnum.DEFAULT;
+    private DataWriterBFEnum dataWriterBfEnum = DataWriterBFEnum.NONE;
     private final int HEADER_H = 14;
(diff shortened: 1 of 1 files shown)


retiutut:
While you are free to use and modify the GUI, this change could negatively impact results for all users. We do want to save both kinds of data by default, in case users decide later to revisit a session and use it in something like Python for advanced processing.

This PR also bypassed the process of creating an issue and discussing whether it should be completed.

bleugreen:
Thanks for the quick review and for clarifying the motivation behind the current default.

Apologies for bypassing the issue process. It was not my intention to skip discussion, but to offer a quick fix for something that seemed inefficient.

I agree that having at least one plain-text export available for post-hoc analysis is important; my concern is just the redundancy and the resulting disk impact. It looks like the BrainFlow `.csv` and GUI `.txt` contain identical data, just in slightly different formats, so users who want to revisit a session can do so with either file. 

I'm just having a hard time understanding why recording the same data in two places should be the default behavior rather than opt-in.

PR #1264, opened 2025-10-31
Fix FFT output over OSC not outputting all bins

### LINKED ISSUE #1263: Fix FFT output over OSC
## Problem

[Forum post](https://openbci.com/forum/index.php?p=/discussion/4072/fft-data-parsing-in-max-msp)

The main problem is that the outputUsingProtocol() method is in the wrong spot.

Image

## Expected 

Proper OSC formatting with all of the data making it across the networking connection

All OS

## GUI Version

6.0.0-beta

Both

## Type of OpenBCI Board

All boards

This screenshot shows that we only receive bin 124.

Image

retiutut:
Fixes #1263

### DIFF
--- a/Networking-Test-Kit/OSC/max-fft-parsing.js
+++ b/Networking-Test-Kit/OSC/max-fft-parsing.js
@@ -1,0 +1,208 @@
+// OpenBCI FFT Data Parser for Max/MSP
+// PROVIDED AS-IS UNDER MIT LICENSE. NO WARRANTY OF ANY KIND. USE AT YOUR OWN RISK.
+// Usage:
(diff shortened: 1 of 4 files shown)


retiutut:
The OSC section of the Networking Output doc seems to have not been updated

https://docs.google.com/document/d/e/2PACX-1vR_4DXPTh1nuiOwWKwIZN3NkGP3kRwpP4Hu6fQmy3jRAOaydOuEI1jket6V4V6PG4yIG15H1N7oFfdV/pub

This doesn't reflect the new best practice of sending a single value to more specific addresses.

